PR #6, opened 2020-04-07
Adding AUTH support with PLAIN, LOGIN and XOAUTH2 mechanisms

OnePunMan:
## Description
Adding the ability to authenticate using the *PLAIN* and *LOGIN* mechanisms.

## Motivation and Context
Authentication is necessary to be able to send emails.
- The *PLAIN* mechanism is one of the most commonly used methods. ([RFC Reference](https://tools.ietf.org/html/rfc4616))
- The *LOGIN* mechanism, although older, is still used by many older clients and servers
- The *XOAUTH2* mechanism is a commonly used OAuth 2.0 to authenticate  ([RFC Reference](https://tools.ietf.org/html/rfc7628))

## How Has This Been Tested?
- Unit tests with dummy data to check for correct encoding, formatting, and logging (100% Coverage)
- Checking debugging data to ensure that sensitive secrets are not logged (placeholder text is used)
- Tests with real accounts (Gmail, Yahoo) to verify that correct credentials work
- Tests with real accounts (Gmail, Yahoo) to verify that incorrect credentials do not work

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Major release (change is NOT backward compatible with prior release)

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #11, opened 2020-06-25
Adding AUTH support with PLAIN, LOGIN and XOAUTH2 mechanisms

cifer828:
## Description

Take over [AUTH PR#6](https://github.com/yahoo/smtp-client-nio/pull/6/commits), which added the ability to authenticate using the PLAIN/LOGIN/XOAUTH2 mechanisms, and address comments.

## Motivation and Context

From [AUTH PR#6](https://github.com/yahoo/smtp-client-nio/pull/6/commits)
> Authentication is necessary to be able to send emails.
> 
> The PLAIN mechanism is one of the most commonly used methods. [(RFC Reference)](https://tools.ietf.org/html/rfc4616)
> The LOGIN mechanism, although older, is still used by many older clients and servers
> The XOAUTH2 mechanism is a commonly used OAuth 2.0 to authenticate [(RFC Reference)](https://tools.ietf.org/html/rfc7628)

## How Has This Been Tested?

- Use existing unit test. Make some change to AuthenticationLoginCommandTest to test if exceptions are caught correctly.
- Test AUTH LOGIN/ PLAIN and OXAUTH2 with real Gmail/Yahoo/Aol accounts. 
- Outlook only support starttls over port 587/25,  so it needs a different connection flow via starttls. Will be solved in another PR. 
- Refer to [smtp-nio test](https://docs.google.com/spreadsheets/d/1XLhJ6giVsU2FIBIN1MPFC88YDW8XsxN6QneMiCKUEgM/edit#gid=0) for more details. 

## Types of changes

- [ ] New feature (non-breaking change which adds functionality)

## Checklist:

### DIFF
--- a/core/src/main/java/com/yahoo/smtpnio/async/request/AbstractAuthenticationCommand.java
+++ b/core/src/main/java/com/yahoo/smtpnio/async/request/AbstractAuthenticationCommand.java
@@ -16,7 +16,6 @@ public abstract class AbstractAuthenticationCommand extends AbstractSmtpCommand 
     /** String literal for "AUTH". */
     protected static final String AUTH = "AUTH";
 
-
     /** Byte array for "AUTH". */
     protected static final byte[] AUTH_B = AUTH.getBytes(StandardCharsets.US_ASCII);
 
@@ -43,7 +42,7 @@ public abstract class AbstractAuthenticationCommand extends AbstractSmtpCommand 
      */
     protected AbstractAuthenticationCommand(@Nonnull final Mechanism mechanism) {
         super(AUTH);
-        this.mechanism = mechanism.toString();
+        this.mechanism = mechanism.name();
     }
 
     /**
--- a/core/src/main/java/com/yahoo/smtpnio/async/request/AbstractAuthenticationCommand.java
+++ b/core/src/main/java/com/yahoo/smtpnio/async/request/AbstractAuthenticationCommand.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright Verizon Media
+ * Licensed under the terms of the Apache 2.0 license. See LICENSE file in project root for terms.
+ */
+package com.yahoo.smtpnio.async.request;
+
+import java.nio.charset.StandardCharsets;
+
+import javax.annotation.Nonnull;
+
+/**
+ * This is the base class for all Authentication (AUTH) commands.
+ */
+public abstract class AbstractAuthenticationCommand extends AbstractSmtpCommand {
+
+    /** String literal for "AUTH". */
+    protected static final String AUTH = "AUTH";
+
+
+    /** Byte array for "AUTH". */
+    protected static final byte[] AUTH_B = AUTH.getBytes(StandardCharsets.US_ASCII);
+
+    /** Mechanism used for authentication. */
+    protected String mechanism;
+
+    /** Three AUTH mechanisms. */
+    protected enum Mechanism {
+        /** AUTH LOGIN. */
+        LOGIN,
+
+        /** AUTH PLAIN. */
+        PLAIN,
+
+        /** AUTH XOAUTH2. */
+        XOAUTH2
+    }
+
+    /**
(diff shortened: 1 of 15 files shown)


core/src/main/java/com/yahoo/smtpnio/async/request/SmtpRFCSupportedCommandType.java
```diff
@@ -16,5 +16,16 @@
...
 
     /** The Quit command (QUIT). */
-    QUIT
+    QUIT,
+
+    /** The Authentication command (AUTH). */
+    AUTH,
+
+    /** The Starttls command (STARTTLS). */
+    STARTTLS;
```
fansu, line 25: remove STARTTLS, we have not supported it yet,

core/src/main/java/com/yahoo/smtpnio/async/request/AbstractAuthenticationCommand.java
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * Initializes an AUTH command object used to authenticate via a specified SASL mechanism that cannot be completed in one command
+     * and requires additional challenge responses from the server.
+     *
+     * @param mechanism authentication mechanism
+     */
+    protected AbstractAuthenticationCommand(@Nonnull final Mechanism mechanism) {
+        super(AUTH);
+        this.mechanism = mechanism.toString();
```
fansu, line 45: lets do .name() instead of toString()

core/src/main/java/com/yahoo/smtpnio/async/request/AuthenticationPlainCommand.java
```diff
@@ -0,0 +1,73 @@
...
+
+
+    /**
+     * Initializes an AUTH command to authenticate via plaintext. This constructor will encode the authzid, username and password into base64.
+     *
+     * @param authorizationIdentity The authorization identity (aka. authzid)
+     * @param username username/authentication identity of the intended sender, usually an email address, in clear text (aka. authcid)
+     * @param password password associated with the above username, in clear text
+     */
+    public AuthenticationPlainCommand(final String authorizationIdentity, @Nonnull final String username, @Nonnull final String password) {
```
fansu, line 38: @Nullable final String authorizationIdentity

core/src/main/java/com/yahoo/smtpnio/async/request/AuthenticationPlainCommand.java
```diff
@@ -0,0 +1,73 @@
...
+        super(Mechanism.PLAIN);
+        final String authzid = authorizationIdentity == null ? "" : authorizationIdentity;
+        this.secret = Base64.encodeBase64((authzid + SmtpClientConstants.NULL + username + SmtpClientConstants.NULL + password)
```
fansu, line 41: please use string builder to construct authzid + SmtpClientConstants.NULL + username + SmtpClientConstants.NULL + password

cifer828: sure. also replaced String concat in other places.

fansu:
One more thing, can you squash all commit and put the into one?

fansu:
@jui8wang as well

cifer828:
> One more thing, can you squash all commit and put the into one?

Combine all commits into two, one for Jackie's AUTH code, one for my debug.

cifer828:
The reason why Gmail needs SNI is that JdkSslContext in old Netty doesn't support the cipher Gmail smtp server uses for TLS. Another interesting finding is that if we use the old Netty for Gmail, all it needs is an "SNI", even if that SNI is incorrect(e.g. "xxx.xxx.xxx" also works, but SNi can't be as same as the server host). I think one possible explanation may be that, by providing SNI, we trigger some downgrading mechanism on the gmail server, but more research are needed for more insights. Anyway, bumping Netty version fixes the issue.

cifer828:
> can you see if 4.1.48.Final works. this is same version we are running in production now.

Yes. 4.1.48.final also works.

PR #13, opened 2020-06-30
User 3 handlers to implement starttls flow

cifer828:
## Description

Current working flow.
- If ```enableSsl``` is true, initialize connection with ```SslHandler``` and ```SslDetectHandler```.
- If ```SslDetectHandler``` works fine, add initializer handlers as normal. If ```SslDetectHandler``` catches```NotSslRecordException``` and starttls is enabled according to session config, close current channel and create a new one without ssl. Add ```PlainReconnectGreetingHandler``` to start the Starttls flow.
- Starttls flow
   1. ```PlainReconnectGreetingHandler``` - receives greeting from server after plain re-connection and send EHLO command. Replace itself by``` StarttlsEhloHandler```.
   2. ```StarttlsEhloHandler```  - receives EHLO response and send STARTTLS. Replace itself by``` StarttlsEhloHandler```.
   3. ```StarttlsSessionHandler```  - receives STARTLS response and add SslHandler. Session is done here.

## Motivation and Context

STARTTLS is required for Outlook connection, since Outlook smtp server doesn't support general SSL/TLS connection. 

## How Has This Been Tested?

Prototype for review, no UTs added.

## Types of changes

## Checklist:

- [ ] I have read the **CONTRIBUTING** document.

### DIFF
--- a/core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
+++ b/core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
@@ -6,6 +6,7 @@ package com.yahoo.smtpnio.async.client;
 
 import java.net.InetSocketAddress;
 import java.util.Collection;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
@@ -19,6 +20,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.yahoo.smtpnio.async.exception.SmtpAsyncClientException;
+import com.yahoo.smtpnio.async.exception.SmtpAsyncClientException.FailureType;
 import com.yahoo.smtpnio.async.netty.PlainReconnectGreetingHandler;
 import com.yahoo.smtpnio.async.netty.SmtpClientConnectHandler;
 import com.yahoo.smtpnio.async.netty.SslHandlerBuilder;
@@ -31,10 +33,12 @@ import io.netty.channel.ChannelPipeline;
 import io.netty.channel.EventLoopGroup;
 import io.netty.channel.nio.NioEventLoopGroup;
 import io.netty.channel.socket.nio.NioSocketChannel;
+import io.netty.handler.ssl.NotSslRecordException;
 import io.netty.handler.ssl.SslContext;
 import io.netty.handler.ssl.SslContextBuilder;
 import io.netty.handler.ssl.SslHandler;
 import io.netty.util.concurrent.GenericFutureListener;
+import io.netty.util.concurrent.Promise;
 
 /**
  * Implementation for the SMTP NIO client.
@@ -64,6 +68,9 @@ public class SmtpAsyncClient {
     public static final String RE_CONNECTION_REC = "[{},{}] try starttls. Re-connecting with non-ssl. "
             + "host={}, port={}, sslEnabled={}, startTlsEnabled={}, sniNames={}";
 
+    /** Handler name for idle sate handler. */
+    private static final String IDLE_STATE_HANDLER_NAME = "idleStateHandler";
+
     /** The SSL context. */
     private final SslContext sslContext;
 
@@ -79,6 +86,9 @@ public class SmtpAsyncClient {
 
     /** Event loop group that will serve all channels for the SMTP client. */
     private final EventLoopGroup group;
+
+    /** This SmtpAsyncClient object . */
+    private final SmtpAsyncClient smtpAsyncClient;
 
     /**
      * Constructs a NIO-based SMTP client.
@@ -103,6 +113,7 @@ public class SmtpAsyncClient {
         this.logger = logger;
         this.bootstrap = bootstrap;
         this.group = group;
+        this.smtpAsyncClient = this;
         bootstrap.channel(NioSocketChannel.class);
         bootstrap.group(group);
     }
@@ -160,14 +171,7 @@ public class SmtpAsyncClient {
         final Collection<String> sniNames = sessionData.getSniNames();
         final Object sessionCtx = sessionData.getSessionContext();
 
-        if (isSsl) {
-            // only add sslHandler for the initial ssl connection
-            bootstrap.handler(new SslDetectHandler(sessionCount.get(), sessionData, config, LoggerFactory.getLogger(SslDetectHandler.class),
-                    debugOption, this, sessionCreatedFuture));
-        } else {
-            // for reconnection and initial non-ssl connection, using plain initializer
-            bootstrap.handler(new SmtpClientChannelInitializer(config.getReadTimeout(), TimeUnit.MILLISECONDS));
-        }
+        bootstrap.handler(new SmtpClientChannelInitializer(config.getReadTimeout(), TimeUnit.MILLISECONDS));
 
         final ChannelFuture nettyConnectFuture = localAddress == null ? bootstrap.connect(host, port)
                 : bootstrap.connect(new InetSocketAddress(host, port), localAddress);
@@ -202,9 +206,38 @@ public class SmtpAsyncClient {
                 // add session specific handlers
                 switch (sessionMode) {
                 case SSL:
-                    // add sslHandler for initial ssl connection
+                    // create sslHandler for ssl connection
                     final SslHandler sslHandler = SslHandlerBuilder.newBuilder(sslContext, ch.alloc(), host, port, sniNames).build();
-                    pipeline.addFirst(SSL_HANDLER, sslHandler);
+                    Promise<Channel> sslConnectionFture = (Promise<Channel>) sslHandler.handshakeFuture();
+                    // check if ssl connection succeed
+                    sslConnectionFture.addListener(new GenericFutureListener<io.netty.util.concurrent.Future<? super Channel>>() {
+                        @Override
+                        public void operationComplete(io.netty.util.concurrent.Future<? super Channel> future) throws Exception {
+                            if (!future.isSuccess()) {
+                                ch.close();
+                                // fail early if ssl connection can't established
+                                try {
+                                    future.get();
+                                } catch (ExecutionException e) {
+                                    if (e.getCause() instanceof NotSslRecordException && isStarttls) {
+                                        // do re-connection with starttls, shouldn't fail here.
+                                        return;
+                                    }
+                                }
+                                SmtpAsyncClientException ex = new SmtpAsyncClientException(
+                                        SmtpAsyncClientException.FailureType.CONNECTION_FAILED_EXCEPTION, future.cause());
+                                logger.error(CONNECT_RESULT_REC, sessionId, sessionCtx, "failure", host, port, isSsl, sniNames, ex);
+                                sessionCreatedFuture
+                                        .done(new SmtpAsyncClientException(FailureType.CONNECTION_FAILED_EXCEPTION, sessionId, sessionCtx));
+                            }
+                        }
+                    });
+                    // pipeline: IdleHandler, SslHandler, SslDetectHandler, other initializer handlers, SmtpClientConnectHandler
+                    pipeline.addAfter(IDLE_STATE_HANDLER_NAME, SslDetectHandler.HANDLER_NAME, new SslDetectHandler(sessionCount.get(), sessionData,
+                            config, LoggerFactory.getLogger(SslDetectHandler.class), debugOption, smtpAsyncClient, sessionCreatedFuture));
+                    pipeline.addBefore(SslDetectHandler.HANDLER_NAME, SSL_HANDLER, sslHandler);
+                    pipeline.addLast(SmtpClientConnectHandler.HANDLER_NAME, new SmtpClientConnectHandler(sessionCreatedFuture,
+                            LoggerFactory.getLogger(SmtpClientConnectHandler.class), debugOption, sessionId, sessionCtx));
                     break;
                 case STARTTLS:
                     // add PlainGreetingHandler to start starttls flow during re-connection
--- a/core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
+++ b/core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
@@ -9,6 +9,7 @@ import java.util.Collection;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.LongUnaryOperator;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
@@ -185,7 +186,13 @@ public class SmtpAsyncClient {
                     return;
                 }
 
-                final long sessionId = sessionCount.getAndIncrement();
+                // avoid negative ids
+                final long sessionId = sessionCount.getAndUpdate(new LongUnaryOperator() { // atomic update
+                    @Override
+                    public long applyAsLong(final long counter) { // increment by 1 unless it overflows
+                        return counter + 1 < 0 ? 1 : counter + 1;
+                    }
+                });
 
                 final Channel ch = nettyConnectFuture.channel();
                 final ChannelPipeline pipeline = ch.pipeline();
--- a/core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
+++ b/core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
@@ -11,7 +11,6 @@ import java.util.List;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
-import java.util.function.LongUnaryOperator;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
@@ -26,6 +25,7 @@ import org.slf4j.LoggerFactory;
 
 import com.yahoo.smtpnio.async.exception.SmtpAsyncClientException;
 import com.yahoo.smtpnio.async.internal.SmtpAsyncSessionImpl;
+import com.yahoo.smtpnio.async.netty.PlainGreetingHandler;
 import com.yahoo.smtpnio.async.netty.SmtpClientConnectHandler;
 
 import io.netty.bootstrap.Bootstrap;
@@ -51,6 +51,10 @@ public class SmtpAsyncClient {
 
     /** Debug record string template. */
     public static final String CONNECT_RESULT_REC = "[{},{}] connect operation complete. result={}, host={}, port={}, sslEnabled={}, sniNames={}";
+
+    /** Debug record string template for reconnection. */
+    public static final String RE_CONNECTION_REC = "[{},{}] try starttls. Re-connecting with non-ssl. "
+            + "host={}, port={}, sslEnabled={}, startTlsEnabled={}, sniNames={}";
 
     /** The SSL context. */
     private final SslContext sslContext;
@@ -106,7 +110,37 @@ public class SmtpAsyncClient {
     public Future<SmtpAsyncCreateSessionResponse> createSession(@Nonnull final SmtpAsyncSessionData sessionData,
             @Nonnull final SmtpAsyncSessionConfig config, @Nonnull final SmtpAsyncSession.DebugMode debugOption) {
         bootstrap.option(ChannelOption.CONNECT_TIMEOUT_MILLIS, config.getConnectionTimeout());
+        final SmtpFuture<SmtpAsyncCreateSessionResponse> sessionCreatedFuture = new SmtpFuture<>();
+        createSession(sessionData, config, debugOption, sessionCreatedFuture, false);
+        return sessionCreatedFuture;
+    }
 
+    /**
+     * Re-connect to the remote server using plain connection and start the Starttls flow.
+     *
+     * @param sessionData a {@link SmtpAsyncSessionData} object containing information about the connection
+     * @param config configuration to be used for this session/connection
+     * @param debugOption the debugging option used
+     * @param sessionCreatedFuture future containing the result of the request
+     */
+    void createStartTlsSession(@Nonnull final SmtpAsyncSessionData sessionData,
+            @Nonnull final SmtpAsyncSessionConfig config,
+            @Nonnull final SmtpAsyncSession.DebugMode debugOption, @Nonnull final SmtpFuture<SmtpAsyncCreateSessionResponse> sessionCreatedFuture) {
+        createSession(sessionData, config, debugOption, sessionCreatedFuture, true);
+    }
+
+    /**
+     * Connects to the remote server asynchronously according to session state.
+     *
+     * @param sessionData a {@link SmtpAsyncSessionData} object containing information about the connection
+     * @param config configuration to be used for this session/connection
+     * @param debugOption the debugging option used
+     * @param sessionCreatedFuture future containing the result of the request
+     * @param isReconnecting whether it's a reconnection session
+     */
+    private void createSession(@Nonnull final SmtpAsyncSessionData sessionData, @Nonnull final SmtpAsyncSessionConfig config,
+            @Nonnull final SmtpAsyncSession.DebugMode debugOption, @Nonnull final SmtpFuture<SmtpAsyncCreateSessionResponse> sessionCreatedFuture,
+            final boolean isReconnecting) {
         final String host = sessionData.getHost();
 
         if (host == null) {
@@ -118,12 +152,12 @@ public class SmtpAsyncClient {
         final InetSocketAddress localAddress = sessionData.getLocalAddress();
         final Collection<String> sniNames = sessionData.getSniNames();
         final Object sessionCtx = sessionData.getSessionContext();
-        final SmtpFuture<SmtpAsyncCreateSessionResponse> sessionFuture = new SmtpFuture<>();
 
-        // when ssl is enabled, detect if server support smtps. if not, try to use starttls flow by adding SslDetectHandler
-        if (enableSsl) {
-            bootstrap.handler(new SslDetectHandler(bootstrap, sessionCount, sessionData, config, debugOption, sessionFuture));
+        if (!isReconnecting && enableSsl) {
+            // only add sslHandler for the initial ssl connection
+            bootstrap.handler(new SslDetectHandler(sessionCount, sessionData, config, debugOption, this, sessionCreatedFuture));
         } else {
+            // for reconnection and initial non-ssl connection, using plain initializer
             bootstrap.handler(new SmtpClientChannelInitializer(config.getReadTimeout(), TimeUnit.MILLISECONDS));
         }
 
@@ -137,26 +171,20 @@ public class SmtpAsyncClient {
                 if (!future.isSuccess()) {
                     final SmtpAsyncClientException ex = new SmtpAsyncClientException(SmtpAsyncClientException.FailureType.WRITE_TO_SERVER_FAILED,
                             future.cause());
-                    sessionFuture.done(ex);
+                    sessionCreatedFuture.done(ex);
                     logger.error(CONNECT_RESULT_REC, "N/A", sessionCtx, "failure", host, port, enableSsl, sniNames, ex);
                     closeChannel(nettyConnectFuture.channel());
                     return;
                 }
 
-                final long sessionId = sessionCount.getAndUpdate(new LongUnaryOperator() { // atomic update
-                    @Override
-                    public long applyAsLong(final long counter) { // increment by 1 unless it overflows
-                        return counter + 1 < 0 ? 1 : counter + 1;
-                    }
-                });
+                final long sessionId = sessionCount.getAndIncrement();
 
-                // add the session specific handlers
                 final Channel ch = nettyConnectFuture.channel();
                 final ChannelPipeline pipeline = ch.pipeline();
 
-                // SSL/TLS is required in many SMTP servers
-                if (enableSsl) {
-                    final SSLEngine engine = sslContext.newEngine(ch.alloc(), host, port);
+                // add the session specific handlers
+                if (!isReconnecting && enableSsl) {
+                    // add sslHandler for initial ssl connection
                     if (sniNames != null && !sniNames.isEmpty()) { // SNI support
                         final List<SNIServerName> serverNames = new ArrayList<>();
                         for (final String sni : sniNames) {
@@ -164,28 +192,28 @@ public class SmtpAsyncClient {
                         }
                         final SSLParameters params = new SSLParameters();
                         params.setServerNames(serverNames);
+                        final SSLEngine engine = sslContext.newEngine(ch.alloc(), host, port);
                         engine.setSSLParameters(params);
+                        pipeline.addFirst(SSL_HANDLER, new SslHandler(engine)); // in / outbound
+                    } else {
+                        pipeline.addFirst(SSL_HANDLER, sslContext.newHandler(ch.alloc(), host, port));
                     }
-                    pipeline.addFirst(new SslHandler(engine));
+                } else if (isReconnecting) {
+                    // add PlainGreetingHandler for reconnection to start starttls flow
+                    pipeline.addLast(PlainGreetingHandler.HANDLER_NAME, new PlainGreetingHandler(sessionCreatedFuture,
+                            LoggerFactory.getLogger(SmtpAsyncSessionImpl.class), debugOption, sessionId, sessionCtx, sessionData));
                 } else {
-                    pipeline.addLast(SmtpClientConnectHandler.HANDLER_NAME, new SmtpClientConnectHandler(sessionFuture,
+                    // add SmtpClientConnectHandler for initial non-ssl connection
+                    pipeline.addLast(SmtpClientConnectHandler.HANDLER_NAME, new SmtpClientConnectHandler(
+                            sessionCreatedFuture,
                             LoggerFactory.getLogger(SmtpAsyncSessionImpl.class), debugOption, sessionId, sessionCtx));
                 }
-
 
                 if (logger.isTraceEnabled() || debugOption == SmtpAsyncSession.DebugMode.DEBUG_ON) {
                     logger.debug(CONNECT_RESULT_REC, sessionId, sessionCtx, "success", host, port, enableSsl, sniNames);
                 }
             }
         });
-        return sessionFuture;
-    }
-
-    /**
-     * Shuts down the client gracefully.
-     */
-    public void shutdown() {
-        group.shutdownGracefully();
     }
 
     /**
@@ -196,6 +224,13 @@ public class SmtpAsyncClient {
     private void closeChannel(@Nullable final Channel channel) {
         if (channel != null && channel.isActive()) {
             channel.close();
+        }
     }
-}
+
+    /**
+     * Shuts down the client gracefully.
+     */
+    public void shutdown() {
+        group.shutdownGracefully();
+    }
 }
--- a/core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
+++ b/core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
@@ -29,6 +29,7 @@ import com.yahoo.smtpnio.async.netty.PlainGreetingHandler;
 import com.yahoo.smtpnio.async.netty.SmtpClientConnectHandler;
 
 import io.netty.bootstrap.Bootstrap;
+import io.netty.buffer.ByteBufAllocator;
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelFuture;
 import io.netty.channel.ChannelOption;
@@ -185,19 +186,7 @@ public class SmtpAsyncClient {
                 // add the session specific handlers
                 if (!isReconnecting && enableSsl) {
                     // add sslHandler for initial ssl connection
-                    if (sniNames != null && !sniNames.isEmpty()) { // SNI support
-                        final List<SNIServerName> serverNames = new ArrayList<>();
-                        for (final String sni : sniNames) {
-                            serverNames.add(new SNIHostName(sni));
-                        }
-                        final SSLParameters params = new SSLParameters();
-                        params.setServerNames(serverNames);
-                        final SSLEngine engine = sslContext.newEngine(ch.alloc(), host, port);
-                        engine.setSSLParameters(params);
-                        pipeline.addFirst(SSL_HANDLER, new SslHandler(engine)); // in / outbound
-                    } else {
-                        pipeline.addFirst(SSL_HANDLER, sslContext.newHandler(ch.alloc(), host, port));
-                    }
+                    pipeline.addFirst(SSL_HANDLER, newSslHandler(sslContext, ch.alloc(), host, port, sniNames));
                 } else if (isReconnecting) {
                     // add PlainGreetingHandler for reconnection to start starttls flow
                     pipeline.addLast(PlainGreetingHandler.HANDLER_NAME, new PlainGreetingHandler(sessionCreatedFuture,
@@ -214,6 +203,34 @@ public class SmtpAsyncClient {
                 }
             }
         });
+    }
+
+    /**
+     * Create a new {@link SslHanlder} for decryption/encryption.
+     *
+     * @param sslContext SslContext used to create SslHandler
+     * @param alloc Allocator for ByteBuf objects
+     * @param host host name of server
+     * @param port port of server
+     * @param sniNames collection of SNI namess
+     * @return an SslHandler to process ssl connection
+     */
+    public static SslHandler newSslHandler(@Nonnull final SslContext sslContext, @Nonnull final ByteBufAllocator alloc,
+            @Nullable final String host,
+            final int port, @Nullable final Collection<String> sniNames) {
+        if (sniNames != null && !sniNames.isEmpty()) { // SNI support
+            final List<SNIServerName> serverNames = new ArrayList<>();
+            for (final String sni : sniNames) {
+                serverNames.add(new SNIHostName(sni));
+            }
+            final SSLParameters params = new SSLParameters();
+            params.setServerNames(serverNames);
+            final SSLEngine engine = sslContext.newEngine(alloc, host, port);
+            engine.setSSLParameters(params);
+            return new SslHandler(engine);
+        } else {
+            return sslContext.newHandler(alloc, host, port);
+        }
     }
 
     /**
--- a/core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
+++ b/core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
@@ -144,7 +144,7 @@ public class SmtpAsyncClient {
      * @param config configuration to be used for this session/connection
      * @param debugOption the debugging option used
      * @param sessionCreatedFuture future containing the result of the request
-     * @param isSsl whether this session use SSL for connection
+     * @param isSsl whether this session uses SSL for connection
      */
     private void createSession(@Nonnull final SmtpAsyncSessionData sessionData, @Nonnull final SmtpAsyncSessionConfig config,
             @Nonnull final SmtpAsyncSession.DebugMode debugOption, @Nonnull final SmtpFuture<SmtpAsyncCreateSessionResponse> sessionCreatedFuture,
@@ -209,6 +209,9 @@ public class SmtpAsyncClient {
                     // add SmtpClientConnectHandler for initial non-ssl connection
                     pipeline.addLast(SmtpClientConnectHandler.HANDLER_NAME, new SmtpClientConnectHandler(sessionCreatedFuture,
                             LoggerFactory.getLogger(SmtpAsyncSessionImpl.class), debugOption, sessionId, sessionCtx));
+                    break;
+                default:
+                    break;
                 }
 
                 if (logger.isTraceEnabled() || debugOption == SmtpAsyncSession.DebugMode.DEBUG_ON) {
--- a/core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
+++ b/core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
@@ -162,7 +162,8 @@ public class SmtpAsyncClient {
 
         if (isSsl) {
             // only add sslHandler for the initial ssl connection
-            bootstrap.handler(new SslDetectHandler(sessionCount.get(), sessionData, config, debugOption, this, sessionCreatedFuture));
+            bootstrap.handler(new SslDetectHandler(sessionCount.get(), sessionData, config, LoggerFactory.getLogger(SslDetectHandler.class),
+                    debugOption, this, sessionCreatedFuture));
         } else {
             // for reconnection and initial non-ssl connection, using plain initializer
             bootstrap.handler(new SmtpClientChannelInitializer(config.getReadTimeout(), TimeUnit.MILLISECONDS));
@@ -196,7 +197,7 @@ public class SmtpAsyncClient {
                 final ChannelPipeline pipeline = ch.pipeline();
 
                 // decide session state
-                SessionMode sessionMode = isSsl ? SessionMode.SSL : (isStarttls ? SessionMode.STARTTLS : SessionMode.NON_SSL);
+                final SessionMode sessionMode = isSsl ? SessionMode.SSL : (isStarttls ? SessionMode.STARTTLS : SessionMode.NON_SSL);
 
                 // add session specific handlers
                 switch (sessionMode) {
--- a/core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
+++ b/core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
@@ -197,7 +197,7 @@ public class SmtpAsyncClient {
                 switch (sessionMode) {
                 case SSL:
                     // add sslHandler for initial ssl connection
-                    final SslHandler sslHandler = new SslHandlerBuilder(sslContext, ch.alloc(), host, port, sniNames).build();
+                    final SslHandler sslHandler = SslHandlerBuilder.newBuilder(sslContext, ch.alloc(), host, port, sniNames).build();
                     pipeline.addFirst(SSL_HANDLER, sslHandler);
                     break;
                 case STARTTLS:
--- a/core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
+++ b/core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
@@ -19,7 +19,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.yahoo.smtpnio.async.exception.SmtpAsyncClientException;
-import com.yahoo.smtpnio.async.internal.SmtpAsyncSessionImpl;
 import com.yahoo.smtpnio.async.netty.PlainReconnectGreetingHandler;
 import com.yahoo.smtpnio.async.netty.SmtpClientConnectHandler;
 import com.yahoo.smtpnio.async.netty.SslHandlerBuilder;
@@ -132,8 +131,7 @@ public class SmtpAsyncClient {
      * @param debugOption the debugging option used
      * @param sessionCreatedFuture future containing the result of the request
      */
-    void createStarttlsSession(@Nonnull final SmtpAsyncSessionData sessionData,
-            @Nonnull final SmtpAsyncSessionConfig config,
+    void createStarttlsSession(@Nonnull final SmtpAsyncSessionData sessionData, @Nonnull final SmtpAsyncSessionConfig config,
             @Nonnull final SmtpAsyncSession.DebugMode debugOption, @Nonnull final SmtpFuture<SmtpAsyncCreateSessionResponse> sessionCreatedFuture) {
         createSession(sessionData, config, debugOption, sessionCreatedFuture, false);
     }
@@ -200,7 +198,7 @@ public class SmtpAsyncClient {
                 // decide session state
                 SessionMode sessionMode = isSsl ? SessionMode.SSL : (isStarttls ? SessionMode.STARTTLS : SessionMode.NON_SSL);
 
-                // add the session specific handlers
+                // add session specific handlers
                 switch (sessionMode) {
                 case SSL:
                     // add sslHandler for initial ssl connection
@@ -210,12 +208,12 @@ public class SmtpAsyncClient {
                 case STARTTLS:
                     // add PlainGreetingHandler to start starttls flow during re-connection
                     pipeline.addLast(PlainReconnectGreetingHandler.HANDLER_NAME, new PlainReconnectGreetingHandler(sessionCreatedFuture,
-                            LoggerFactory.getLogger(SmtpAsyncSessionImpl.class), debugOption, sessionId, sessionCtx, sessionData));
+                            LoggerFactory.getLogger(PlainReconnectGreetingHandler.class), debugOption, sessionId, sessionData));
                     break;
                 case NON_SSL:
                     // add SmtpClientConnectHandler for initial non-ssl connection
                     pipeline.addLast(SmtpClientConnectHandler.HANDLER_NAME, new SmtpClientConnectHandler(sessionCreatedFuture,
-                            LoggerFactory.getLogger(SmtpAsyncSessionImpl.class), debugOption, sessionId, sessionCtx));
+                            LoggerFactory.getLogger(SmtpClientConnectHandler.class), debugOption, sessionId, sessionCtx));
                     break;
                 default:
                     break;
--- a/core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
+++ b/core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
@@ -5,31 +5,25 @@
 package com.yahoo.smtpnio.async.client;
 
 import java.net.InetSocketAddress;
-import java.util.ArrayList;
 import java.util.Collection;
-import java.util.List;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
-import javax.net.ssl.SNIHostName;
-import javax.net.ssl.SNIServerName;
-import javax.net.ssl.SSLEngine;
 import javax.net.ssl.SSLException;
-import javax.net.ssl.SSLParameters;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.yahoo.smtpnio.async.exception.SmtpAsyncClientException;
 import com.yahoo.smtpnio.async.internal.SmtpAsyncSessionImpl;
-import com.yahoo.smtpnio.async.netty.PlainGreetingHandler;
+import com.yahoo.smtpnio.async.netty.PlainReconnectGreetingHandler;
 import com.yahoo.smtpnio.async.netty.SmtpClientConnectHandler;
+import com.yahoo.smtpnio.async.netty.SslHandlerBuilder;
 
 import io.netty.bootstrap.Bootstrap;
-import io.netty.buffer.ByteBufAllocator;
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelFuture;
 import io.netty.channel.ChannelOption;
@@ -46,6 +40,19 @@ import io.netty.util.concurrent.GenericFutureListener;
  * Implementation for the SMTP NIO client.
  */
 public class SmtpAsyncClient {
+
+    /** Enum used to specify different session mode. */
+    private enum SessionMode {
+
+        /** Session will be built for the first time and via SSL connection. */
+        SSL,
+
+        /** Session will be built for the first time and via plain connection. */
+        NON_SSL,
+
+        /** Session will be built as a re-connection and via plain connection. */
+        STARTTLS;
+    }
 
     /** Handler name for the ssl handler. */
     public static final String SSL_HANDLER = "sslHandler";
@@ -112,7 +119,7 @@ public class SmtpAsyncClient {
             @Nonnull final SmtpAsyncSessionConfig config, @Nonnull final SmtpAsyncSession.DebugMode debugOption) {
         bootstrap.option(ChannelOption.CONNECT_TIMEOUT_MILLIS, config.getConnectionTimeout());
         final SmtpFuture<SmtpAsyncCreateSessionResponse> sessionCreatedFuture = new SmtpFuture<>();
-        createSession(sessionData, config, debugOption, sessionCreatedFuture, false);
+        createSession(sessionData, config, debugOption, sessionCreatedFuture, sessionData.isSslEnabled());
         return sessionCreatedFuture;
     }
 
@@ -124,10 +131,10 @@ public class SmtpAsyncClient {
      * @param debugOption the debugging option used
      * @param sessionCreatedFuture future containing the result of the request
      */
-    void createStartTlsSession(@Nonnull final SmtpAsyncSessionData sessionData,
+    void createStarttlsSession(@Nonnull final SmtpAsyncSessionData sessionData,
             @Nonnull final SmtpAsyncSessionConfig config,
             @Nonnull final SmtpAsyncSession.DebugMode debugOption, @Nonnull final SmtpFuture<SmtpAsyncCreateSessionResponse> sessionCreatedFuture) {
-        createSession(sessionData, config, debugOption, sessionCreatedFuture, true);
+        createSession(sessionData, config, debugOption, sessionCreatedFuture, false);
     }
 
     /**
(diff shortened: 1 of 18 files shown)


core/src/main/java/com/yahoo/smtpnio/async/netty/SmtpClientConnectHandler.java
```diff
@@ -89,7 +123,50 @@ public void decode(@Nonnull final ChannelHandlerContext ctx, @Nonnull final Smtp
     @Override
     public void exceptionCaught(@Nonnull final ChannelHandlerContext ctx, @Nonnull final Throwable cause) {
         logger.error("[{},{}] Connection failed due to encountering exception:{}.", sessionId, sessionCtx, cause);
-        sessionCreatedFuture.done(new SmtpAsyncClientException(FailureType.CONNECTION_FAILED_EXCEPTION, cause, sessionId, sessionCtx));
+        // NotSslRecordException is expected during reconnection
+        if (cause.getCause() instanceof NotSslRecordException) {
+            if (logger.isTraceEnabled() || logOpt == SmtpAsyncSession.DebugMode.DEBUG_ON) {
```
nituyahoo, line 128: we need to prevent loops, as the next connection might again throw notsslrecordexception, we want to throw exception after 1 retry

core/src/main/java/com/yahoo/smtpnio/async/netty/SmtpClientConnectHandler.java
```diff
@@ -89,7 +123,50 @@ public void decode(@Nonnull final ChannelHandlerContext ctx, @Nonnull final Smtp
     @Override
     public void exceptionCaught(@Nonnull final ChannelHandlerContext ctx, @Nonnull final Throwable cause) {
         logger.error("[{},{}] Connection failed due to encountering exception:{}.", sessionId, sessionCtx, cause);
-        sessionCreatedFuture.done(new SmtpAsyncClientException(FailureType.CONNECTION_FAILED_EXCEPTION, cause, sessionId, sessionCtx));
+        // NotSslRecordException is expected during reconnection
+        if (cause.getCause() instanceof NotSslRecordException) {
```
nituyahoo, line 127: please add another config value that the client needs to specify saying if they want the nio client to try starttls

core/src/main/java/com/yahoo/smtpnio/async/netty/SmtpClientConnectHandler.java
```diff
@@ -89,7 +123,50 @@ public void decode(@Nonnull final ChannelHandlerContext ctx, @Nonnull final Smtp
...
+            if (logger.isTraceEnabled() || logOpt == SmtpAsyncSession.DebugMode.DEBUG_ON) {
+                logger.debug("Fail to connect to server. Try to reconnect with non-ssl.");
+            }
+
+            // reconnect the channel
+            final String host = sessionData.getHost();
```
nituyahoo, line 133: we do not want to redo the same code that is in the session, please work with Ray on how to combine the two code

core/src/main/java/com/yahoo/smtpnio/async/netty/SmtpClientConnectHandler.java
```diff
@@ -89,7 +123,50 @@ public void decode(@Nonnull final ChannelHandlerContext ctx, @Nonnull final Smtp
...
+                    ch.pipeline().addLast(SmtpClientConnectHandler.HANDLER_NAME, new SmtpClientConnectHandler(sessionCreatedFuture,
+                            LoggerFactory.getLogger(SmtpAsyncSessionImpl.class), logOpt, sessionId, sessionCtx, bootstrap, sessionData));
+                    if (logger.isTraceEnabled() || logOpt == SmtpAsyncSession.DebugMode.DEBUG_ON) {
+                        logger.debug(CONNECT_RESULT_REC, sessionId, sessionCtx, "success", host, port, enableSsl, sniNames);
+                    }
+
+                }
+            });
+            // fire inactive() to disable and unregister current handlers
+            isReconnecting = true;
```
nituyahoo, line 165: why do we need this variable, we should just close the channel, let us not use inactive here

core/src/main/java/com/yahoo/smtpnio/async/netty/SmtpClientConnectHandler.java
```diff
@@ -74,7 +107,8 @@ public void decode(@Nonnull final ChannelHandlerContext ctx, @Nonnull final Smtp
 
         if (serverResponse.getCode().value() == SmtpResponse.Code.GREETING) { // successful response
             // add the command response handler
-            final SmtpAsyncSessionImpl session = new SmtpAsyncSessionImpl(ctx.channel(), logger, logOpt, sessionId, pipeline, sessionCtx);
+            final SmtpAsyncSessionImpl session = new SmtpAsyncSessionImpl(ctx.channel(), logger, logOpt, sessionId, pipeline, sessionCtx,
+                    sessionData);
             final SmtpAsyncCreateSessionResponse response = new SmtpAsyncCreateSessionResponse(session, serverResponse);
```
nituyahoo, line 112: the future cannot be completed here for the startls case as the connection is not yet established

core/src/main/java/com/yahoo/smtpnio/async/internal/SmtpAsyncSessionImpl.java
```diff
@@ -389,11 +439,57 @@ public void handleChannelResponse(@Nonnull final SmtpResponse serverResponse) {
...
         // Multiline responses go here and continues. EHLO is an example
     }
 
+    /**
+     * Create a new SslHandler using host, port and SNIO from sessionData.
+     *
+     * @return a new SslHandler
+     * @throws SSLException on ssl error
+     */
+    private SslHandler createSslHandler() throws SSLException {
```
nituyahoo, line 477: let us not duplicate code

core/src/main/java/com/yahoo/smtpnio/async/internal/SmtpAsyncSessionImpl.java
```diff
@@ -389,11 +439,57 @@ public void handleChannelResponse(@Nonnull final SmtpResponse serverResponse) {
...
+                    starttlsState = StarttlsState.STARTTLS_SENT;
+                } else if (command instanceof ExtendedHelloCommand && starttlsState == StarttlsState.SECOND_EHLO_SENT) {
+                    // receive response of second EHLO command in Starttls flow
+                    firstEhloEntryInStarttls.getFuture().done(doneResponse);
+                    starttlsState = StarttlsState.NON_STARTTLS;
+                } else if (command instanceof StarttlsCommand && starttlsState == StarttlsState.STARTTLS_SENT) {
+                    // receive response of STARTTLS command in Starttls flow
+                    Channel channel = channelRef.get();
+                    channel.pipeline().addFirst(createSslHandler());
+                    execute(new ExtendedHelloCommand(((ExtendedHelloCommand) firstEhloEntryInStarttls.cmd).getName()));
```
nituyahoo, line 458: why do we need to send ehlo after starttls

core/src/main/java/com/yahoo/smtpnio/async/internal/SmtpAsyncSessionImpl.java
```diff
@@ -195,15 +234,25 @@ public void operationComplete(@Nonnull final ChannelFuture future) {
...
+            final long sessionId, final ChannelPipeline pipeline, @Nonnull final Object sessionCtx, @Nonnull final SmtpAsyncSessionData sessionData) {
         this.channelRef.set(channel);
         this.logger = logger;
         this.debugModeRef.set(debugMode);
         this.sessionId = sessionId;
         this.requestsQueue = new ConcurrentLinkedQueue<>();
         this.sessionCtx = sessionCtx;
+        this.sessionData = sessionData;
+
+        if (sessionData.isSslEnabled() && channel.pipeline().get(SmtpAsyncClient.SSL_HANDLER) == null) {
```
nituyahoo, line 249: this is not a good condition, you should explicitly pass something to indicate this is starttls flow

core/src/main/java/com/yahoo/smtpnio/async/internal/SmtpAsyncSessionImpl.java
```diff
@@ -389,11 +439,57 @@ public void handleChannelResponse(@Nonnull final SmtpResponse serverResponse) {
             // Completion is indicated when the response does not have a hyphen (-) after the code
             final SmtpAsyncResponse doneResponse = new SmtpAsyncResponse(responses);
             removeFirstEntry();
-            curEntry.getFuture().done(doneResponse);
+
+            try {
+                final SmtpRequest command = curEntry.cmd;
+                if (command instanceof ExtendedHelloCommand && starttlsState == StarttlsState.BEFORE_FIRST_EHLO) {
```
nituyahoo, line 445: why is state not enough, why do we need to check command type

core/src/main/java/com/yahoo/smtpnio/async/internal/SmtpAsyncSessionImpl.java
```diff
@@ -389,11 +439,57 @@ public void handleChannelResponse(@Nonnull final SmtpResponse serverResponse) {
...
+                    // receive response of first EHLO command in Starttls flow
+                    firstEhloEntryInStarttls = curEntry;
+                    execute(new StarttlsCommand());
+                    starttlsState = StarttlsState.STARTTLS_SENT;
+                } else if (command instanceof ExtendedHelloCommand && starttlsState == StarttlsState.SECOND_EHLO_SENT) {
```
nituyahoo, line 450: why do we need second ehlo

core/src/main/java/com/yahoo/smtpnio/async/netty/SmtpClientConnectHandler.java
cifer828: New handlers are used to prevent loop in latest commit.

core/src/main/java/com/yahoo/smtpnio/async/netty/SmtpClientConnectHandler.java
cifer828: Added enableStarttls in SmtpAsyncSessionData.

cifer828:
Using multiple dynamical handlers to implement Starttls flow instead of putting everything into one connection handler. Please see updated description for details.

cifer828:
@fansu @jui8wang

core/src/main/java/com/yahoo/smtpnio/async/netty/EhloHandler.java
```diff
@@ -0,0 +1,146 @@
...
+        if (serverResponse.isLastLineResponse()) {
+            Channel channel = ctx.channel();
+            channel.writeAndFlush(new StarttlsCommand().getCommandLineBytes());
+            ctx.pipeline().replace(this, "starttlsHandler",
+                    new StarttlsHandler(sessionCreatedFuture, logger, logOpt, sessionId, sessionCtx, sessionData));
+            if (logger.isTraceEnabled() || logOpt == SmtpAsyncSession.DebugMode.DEBUG_ON) {
+                logger.debug("[{},{}] EHLO response after reconnection was successful. Trying to sent STARTTLS.", sessionId, sessionCtx);
+            }
+            cleanup();
+        }
```
fansu, line 89: check for bad response

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
```diff
@@ -0,0 +1,201 @@
...
+    @Override
+    public void decode(@Nonnull final ChannelHandlerContext ctx, @Nonnull final SmtpResponse serverResponse, @Nonnull final List<Object> out) {
+        final ChannelPipeline pipeline = ctx.pipeline();
+        pipeline.remove(this);
+
+        if (serverResponse.getCode().value() == SmtpResponse.Code.GREETING) { // successful response
+            // add the command response handler
+            final SmtpAsyncSessionImpl session = new SmtpAsyncSessionImpl(ctx.channel(), logger, logOpt, sessionId, pipeline, sessionCtx);
+            final SmtpAsyncCreateSessionResponse response = new SmtpAsyncCreateSessionResponse(session, serverResponse);
+            sessionCreatedFuture.done(response);
```
fansu, line 99: you cant give done() with success here. It needs to be after below logic. If there is following errors, client assumes it is success and also your sessionCreatedFuture already completed

core/src/main/java/com/yahoo/smtpnio/async/request/StarttlsCommand.java
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * This class defines a STARTTLS (STARTTLS) command.
+ */
+public class StarttlsCommand extends AbstractSmtpCommand {
+
+    /** String literal for "STARTTLS". */
+    private static final String STARTTLS = "STARTTLS";
+
+    /**
+     * Initializes a QUIT command object used to log out of the server.
```
fansu, line 16: STARTTLS command

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
```diff
@@ -137,44 +142,52 @@ public void operationComplete(final io.netty.util.concurrent.Future<? super Void
                     closeChannel(nettyConnectFuture.channel());
                     return;
                 }
+
+                final long sessionId = sessionCount.getAndUpdate(new LongUnaryOperator() { // atomic update
+                    @Override
+                    public long applyAsLong(final long counter) { // increment by 1 unless it overflows
```
jui8wang, line 148: do we have to do this?  can we use incrementAndGet ?

core/src/main/java/com/yahoo/smtpnio/async/netty/EhloHandler.java
```diff
@@ -0,0 +1,146 @@
...
+import io.netty.handler.timeout.IdleStateEvent;
+
+/**
+ * This handler receives EHLO response sent by {@link PlainGreetingHandler}. If the response code if 2xx, send STARTTLS command to notify server for
+ * starting encryption connection.
+ */
+public class EhloHandler extends MessageToMessageDecoder<SmtpResponse> {
+
+    /** Literal for the name registered in pipeline. */
+    public static final String HANDLER_NAME = "SmtpClientConnectHandler";
```
fansu, line 37: fix this handler name

core/src/main/java/com/yahoo/smtpnio/async/netty/EhloHandler.java
```diff
@@ -0,0 +1,146 @@
...
+        this.sessionId = sessionId;
+        this.sessionData = sessionData;
+    }
+
+    @Override
+    public void decode(@Nonnull final ChannelHandlerContext ctx, @Nonnull final SmtpResponse serverResponse, @Nonnull final List<Object> out) {
+        if (serverResponse.isLastLineResponse()) {
+            Channel channel = ctx.channel();
+            channel.writeAndFlush(new StarttlsCommand().getCommandLineBytes());
+            ctx.pipeline().replace(this, "starttlsHandler",
```
fansu, line 83: please use .HANDLER_NAME instead

core/src/main/java/com/yahoo/smtpnio/async/netty/PlainGreetingHandler.java
```diff
@@ -0,0 +1,158 @@
...
+
+    @Override
+    public void decode(@Nonnull final ChannelHandlerContext ctx, @Nonnull final SmtpResponse serverResponse, @Nonnull final List<Object> out) {
+        final ChannelPipeline pipeline = ctx.pipeline();
+        // this handler is solely used to detect connect greeting from server, job done, removing it
+        pipeline.remove(this);
+
+        if (serverResponse.getCode().value() == SmtpResponse.Code.GREETING) { // successful response
+            Channel channel = ctx.channel();
+            channel.writeAndFlush(new ExtendedHelloCommand("local").getCommandLineBytes());
```
fansu, line 87: put "local" into a static variable

core/src/main/java/com/yahoo/smtpnio/async/netty/PlainGreetingHandler.java
```diff
@@ -0,0 +1,158 @@
...
+            ctx.pipeline().addLast(EhloHandler.HANDLER_NAME,
+                    new EhloHandler(sessionCreatedFuture, logger, logOpt, sessionId, sessionCtx, sessionData));
+            if (logger.isTraceEnabled() || logOpt == SmtpAsyncSession.DebugMode.DEBUG_ON) {
+                logger.debug("[{},{}] Server greeting response of reconnection was successful. Starttls flow begins. Sending EHLO.",
+                        sessionId,
+                        sessionCtx);
```
fansu, line 93: fix this line format

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
```diff
@@ -0,0 +1,201 @@
...
+import io.netty.handler.timeout.IdleStateEvent;
+
+/**
+ * This handler receives STARTTLS response sent by {@link EhloHandler}. If the response code if 250, add SslHandler to upgrade current plain
+ * connection to encrypted connection. starting encryption connection.
+ */
+public class StarttlsHandler extends MessageToMessageDecoder<SmtpResponse> {
+
+    /** Literal for the name registered in pipeline. */
+    public static final String HANDLER_NAME = "SmtpClientConnectHandler";
```
fansu, line 48: fix this handler name

core/src/main/java/com/yahoo/smtpnio/async/client/SslDetectHandler.java
```diff
@@ -0,0 +1,240 @@
...
+ * <li>{@link PlainGreetingHandler} processes server greeting and send first EHLO.
+ *
+ * <li>{@link EhloHandler} processes EHLO response and send STARTTLS.
+ *
+ * <li>{@link StarttlsHandler} processes STARTLS response and upgrade plain connection to ssl connection.
+ */
+public class SslDetectHandler extends ByteToMessageDecoder {
+
+    /** Handler name for the ssl handler. */
+    public static final String SSL_DETECT_HANDLER = "SslDetectHandler";
```
fansu, line 52: where do you use this variable, if not, remove.

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
```diff
@@ -137,44 +142,52 @@ public void operationComplete(final io.netty.util.concurrent.Future<? super Void
...
                 // add the session specific handlers
                 final Channel ch = nettyConnectFuture.channel();
                 final ChannelPipeline pipeline = ch.pipeline();
 
                 // SSL/TLS is required in many SMTP servers
                 if (enableSsl) {
+                    final SSLEngine engine = sslContext.newEngine(ch.alloc(), host, port);
                     if (sniNames != null && !sniNames.isEmpty()) { // SNI support
                         final List<SNIServerName> serverNames = new ArrayList<>();
                         for (final String sni : sniNames) {
```
jui8wang, line 162: since u are creating engine earlier now, and new SNIHostName may throw IllegalArgumentException, could u move this populating serverNames out of the addListener implementation.  We do build this serverNames earlier before establishing the connection and if SNI is an invalid host, the exception can be thrown earlier

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
```diff
@@ -137,44 +142,52 @@ public void operationComplete(final io.netty.util.concurrent.Future<? super Void
...
             }
         });
         return sessionFuture;
     }
 
+    /**
+     * Shuts down the client gracefully.
+     */
+    public void shutdown() {
+        group.shutdownGracefully();
```
jui8wang, line 188: lets not move code around if no changes involved, it gives more work for reviewers to see the diff and review.

core/src/main/java/com/yahoo/smtpnio/async/client/SslDetectHandler.java
cifer828: removed.

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncSessionData.java
```diff
@@ -52,6 +55,9 @@
         /** Whether to use SSL or not. */
         private final boolean enableSsl;
 
+        /** Whether to use starttls or not if native smtps failed. */
+        private final boolean enableStarttls;
```
jui8wang, line 59: we do not need this flag, the whole idea that we do this is client only knows the intention that they want to use a secure communication (so your enableSsl flag is the intention).  They do not know whether they can use ssl initially or use startTls, our job is to automatically figuring this out.

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
cifer828: move done() inside try block. only done() if SslHandler is inserted.

core/src/main/java/com/yahoo/smtpnio/async/netty/EhloHandler.java
cifer828: Added response type check. Flow will move forward only if all the responses are 2xx.

core/src/main/java/com/yahoo/smtpnio/async/client/SslDetectHandler.java
```diff
@@ -0,0 +1,240 @@
...
+            ctx.close();
+
+            // if starttls is enbaled, try to create a new connection without ssl
+            if (enableStarttls) {
+                if (logger.isTraceEnabled() || debugOption == SmtpAsyncSession.DebugMode.DEBUG_ON) {
+                    logger.debug(RE_CONNECTION_REC, sessionId, sessionCtx, host, port, false, enableStarttls, sniNames);
+                }
+                bootstrap.handler(new SmtpClientChannelInitializer(sessionConfig.getReadTimeout(), TimeUnit.MILLISECONDS));
+                final ChannelFuture nettyConnectFuture = localAddress == null ? bootstrap.connect(host, port)
+                        : bootstrap.connect(new InetSocketAddress(host, port), localAddress);
```
jui8wang, line 146: In ImapAsyncClient, create a package level (not public) method, createStartTlsSession and move this logic there.  We will start from a separate method (createStartTlsSession) and see if we can make it back to createSession.

core/src/main/java/com/yahoo/smtpnio/async/client/SslDetectHandler.java
```diff
@@ -0,0 +1,240 @@
...
+            }
+        }
+    }
+
+    /**
+     * Closes the connection.
+     *
+     * @param ctx the ChannelHandlerContext
+     */
+    private void close(final ChannelHandlerContext ctx) {
```
jui8wang, line 222: remove, does not belong here

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
cifer828: replaced.

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
cifer828: revert to previous code.

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
cifer828: SNI validation is now checked out of listener.

cifer828: I revert this to the previous version that engine is created after SNIHostName. I tried to move ```populating serverNames``` out of listener, but found that I have to create empty list even if sniNames is empty which is a waste of space. Also, using the previous flow helps combine the code for creating SslHandler.

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
cifer828: revert the previous version.

core/src/main/java/com/yahoo/smtpnio/async/client/SslDetectHandler.java
cifer828: ```createStartTlsSession``` cant call ```createSession``` directly because they intend to add different handlers. I extract the overlap part of initial connection and re-connection, and add a new private overload method  ```createSession```. This new method will be called by ```createStartTlsSession``` and the public```createSession```.

core/src/main/java/com/yahoo/smtpnio/async/client/SslDetectHandler.java
cifer828: I think we may need this method because we want to close channel if we see some unexpected exceptions or events (like what we did to ```SmtpClientConnectHandler``` in the fix close connection PR)

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
```diff
@@ -216,6 +205,34 @@ public void operationComplete(final io.netty.util.concurrent.Future<? super Void
...
+     * Create a new {@link SslHanlder} for decryption/encryption.
+     *
+     * @param sslContext SslContext used to create SslHandler
+     * @param alloc Allocator for ByteBuf objects
+     * @param host host name of server
+     * @param port port of server
+     * @param sniNames collection of SNI namess
+     * @return an SslHandler to process ssl connection
+     */
+    public static SslHandler newSslHandler(@Nonnull final SslContext sslContext, @Nonnull final ByteBufAllocator alloc,
```
cifer828, line 218: Reflect nitu's comment "not duplicate code" on creating SslHandler. not sure if a public static method is a good way to fix that.

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
```diff
@@ -104,11 +109,39 @@ public SmtpAsyncClient(final int numThreads) throws SSLException {
...
+    /**
+     * Connects to the remote server asynchronously according to session state.
+     *
+     * @param sessionData a {@link SmtpAsyncSessionData} object containing information about the connection
+     * @param config configuration to be used for this session/connection
+     * @param debugOption the debugging option used
+     * @param sessionCreatedFuture future containing the result of the request
+     * @param isReconnecting whether it's a reconnection session
+     */
+    private void createSession(@Nonnull final SmtpAsyncSessionData sessionData, @Nonnull final SmtpAsyncSessionConfig config,
```
cifer828, line 142: This private overload method includes overlap part of initial session creation and session re-creation for the sake of reusing code. It will be called by ```createStartTlsSession``` and the public ```createSession```. not sure if this is a proper way.

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
```diff
@@ -121,9 +154,16 @@ public SmtpAsyncClient(final int numThreads) throws SSLException {
         final Collection<String> sniNames = sessionData.getSniNames();
         final Object sessionCtx = sessionData.getSessionContext();
 
-        final SmtpFuture<SmtpAsyncCreateSessionResponse> sessionFuture = new SmtpFuture<>();
-        final ChannelFuture nettyConnectFuture =
-                localAddress == null ? bootstrap.connect(host, port) : bootstrap.connect(new InetSocketAddress(host, port), localAddress);
+        if (!isReconnecting && enableSsl) {
```
fansu, line 157: changing isReconnecting to isSsl; in line 112 do boolean isSsl=sessionData.isSslEnabled()?true:false. remove looking at sessionData.isSslEnabled(); here

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
fansu: Can you have another class SslHandlerBuilder and have constructor, you can create SslHandler with build() function.

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
fansu: looking at the following code after line 186. Can you create private enum in this class called SessionMode (SSL, STARTTLS, NON_SSL) and use sessionData.isSslEnabled() or starttls to decide. Code will be much clear.

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
```diff
@@ -132,47 +172,65 @@ public void operationComplete(final io.netty.util.concurrent.Future<? super Void
...
-                        engine.setSSLParameters(params);
-                        pipeline.addFirst(SSL_HANDLER, new SslHandler(engine)); // in / outbound
-                    } else {
-                        pipeline.addFirst(SSL_HANDLER, sslContext.newHandler(ch.alloc(), host, port));
-                    }
+                // add the session specific handlers
+                if (!isReconnecting && enableSsl) {
+                    // add sslHandler for initial ssl connection
+                    pipeline.addFirst(SSL_HANDLER, newSslHandler(sslContext, ch.alloc(), host, port, sniNames));
+                } else if (isReconnecting) {
```
fansu, line 190: check the comments above

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncSessionData.java
```diff
@@ -25,6 +25,9 @@
     /** Whether to use SSL or not. */
     private final boolean enableSsl;
 
+    /** Whether to use starttls or not if native smtps failed. */
+    private final boolean enableStarttls;
```
fansu, line 29: enableStartTls

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncSessionData.java
fansu: Agree. remove this flag. do not change this class.  client should only able to pass enableSsl flag. Smtp nio library should be able to handle ssl connection either by ssl end-to-end or starttls.

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpClientChannelInitializer.java
```diff
@@ -21,6 +21,9 @@
  */
 final class SmtpClientChannelInitializer extends ChannelInitializer<SocketChannel> {
 
+    /** Literal name for this initializer. */
+    static final String INITIALIZER = "initializer";
```
fansu, line 25: INITIALIZER_NAME

core/src/main/java/com/yahoo/smtpnio/async/netty/PlainGreetingHandler.java
```diff
@@ -0,0 +1,154 @@
...
+import io.netty.channel.Channel;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.handler.codec.MessageToMessageDecoder;
+import io.netty.handler.timeout.IdleState;
+import io.netty.handler.timeout.IdleStateEvent;
+
+/**
+ * This handler receives greeting from server after plain re-connection. If the response code if 250, start Starttls flow by sending EHLO command.
+ */
+public class PlainGreetingHandler extends MessageToMessageDecoder<SmtpResponse> {
```
fansu, line 33: rename to ReconnectGreetingHandler

core/src/main/java/com/yahoo/smtpnio/async/netty/EhloHandler.java
```diff
@@ -0,0 +1,152 @@
...
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.handler.codec.MessageToMessageDecoder;
+import io.netty.handler.timeout.IdleState;
+import io.netty.handler.timeout.IdleStateEvent;
+
+/**
+ * This handler receives EHLO response sent by {@link PlainGreetingHandler}. If the response code if 2xx, send STARTTLS command to notify server for
+ * starting encryption connection.
+ */
+public class EhloHandler extends MessageToMessageDecoder<SmtpResponse> {
```
fansu, line 34: rename to StarttlsEhloHandler

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
```diff
@@ -0,0 +1,169 @@
...
+import io.netty.handler.ssl.SslContext;
+import io.netty.handler.ssl.SslContextBuilder;
+import io.netty.handler.timeout.IdleState;
+import io.netty.handler.timeout.IdleStateEvent;
+
+/**
+ * This handler receives STARTTLS response sent by {@link EhloHandler}. If the response code if 250, add SslHandler to upgrade current plain
+ * connection to encrypted connection. starting encryption connection.
+ */
+public class StarttlsHandler extends MessageToMessageDecoder<SmtpResponse> {
```
fansu, line 38: StarttlsSessionHandler

core/src/main/java/com/yahoo/smtpnio/async/netty/PlainGreetingHandler.java
```diff
@@ -0,0 +1,154 @@
...
+
+    /** Literal for the name registered in pipeline. */
+    public static final String HANDLER_NAME = "PlainGreetingHandler";
```
fansu, line 36: fix same.

core/src/main/java/com/yahoo/smtpnio/async/netty/PlainGreetingHandler.java
```diff
@@ -0,0 +1,154 @@
...
+    private DebugMode logOpt;
+
+    /** Session Id. */
+    private final long sessionId;
+
+    /** Context for session information. */
+    private Object sessionCtx;
+
+    /**
+     * Initializes {@link SmtpClientConnectHandler} to process ok greeting after connection.
```
fansu, line 60: fix this comment "SmtpClientConnectHandler" to ReconnectGreetingHandler

core/src/main/java/com/yahoo/smtpnio/async/netty/PlainGreetingHandler.java
```diff
@@ -0,0 +1,154 @@
...
+     *
+     * @param sessionFuture SMTP session future
+     * @param logger the {@link Logger} instance for {@link SmtpAsyncSessionImpl}
+     * @param logOpt logging option for the session to be created
+     * @param sessionId the session id
+     * @param sessionCtx context for the session information; it is used used for logging
+     * @param sessionData sessionData object containing information about the connection.
+     */
+    public PlainGreetingHandler(@Nonnull final SmtpFuture<SmtpAsyncCreateSessionResponse> sessionFuture, @Nonnull final Logger logger,
+            @Nonnull final DebugMode logOpt, final long sessionId,
```
fansu, line 70: fix these parameter to be one line or less lines

core/src/main/java/com/yahoo/smtpnio/async/netty/PlainGreetingHandler.java
```diff
@@ -0,0 +1,154 @@
...
+            if (logger.isTraceEnabled() || logOpt == SmtpAsyncSession.DebugMode.DEBUG_ON) {
+                logger.debug("[{},{}] Server greeting response of reconnection was successful. Starttls flow begins. Sending EHLO.", sessionId,
+                        sessionCtx);
+            }
+
+        } else {
+            logger.error("[{},{}] Server greeting response of reconnection was not successful:{}", sessionId, sessionCtx, serverResponse.toString());
+            sessionCreatedFuture.done(new SmtpAsyncClientException(FailureType.CONNECTION_FAILED_INVALID_GREETING_CODE, sessionId, sessionCtx,
+                    serverResponse.toString()));
+            close(ctx); // closing the channel if we r not getting a ok greeting
```
fansu, line 97: we are

core/src/main/java/com/yahoo/smtpnio/async/netty/PlainGreetingHandler.java
```diff
@@ -0,0 +1,154 @@
...
+        }
+        cleanup();
+    }
+
+    @Override
+    public void exceptionCaught(@Nonnull final ChannelHandlerContext ctx, @Nonnull final Throwable cause) {
+        logger.error("[{},{}] Re-connection failed due to encountering exception:{}.", sessionId, sessionCtx, cause);
+        sessionCreatedFuture.done(new SmtpAsyncClientException(FailureType.CONNECTION_FAILED_EXCEPTION, cause, sessionId, sessionCtx));
+        cleanup();
```
fansu, line 106: reverse the order, close(ctx) and then cleanup()

core/src/main/java/com/yahoo/smtpnio/async/netty/PlainGreetingHandler.java
```diff
@@ -0,0 +1,154 @@
...
+
+    @Override
+    public void userEventTriggered(@Nonnull final ChannelHandlerContext ctx, @Nonnull final Object msg) {
+        if (msg instanceof IdleStateEvent) { // Handle the IdleState if needed
+            final IdleStateEvent event = (IdleStateEvent) msg;
+            if (event.state() == IdleState.READER_IDLE) {
+                logger.error("[{},{}] Re-connection failed due to taking longer than configured allowed time.", sessionId, sessionCtx);
+                sessionCreatedFuture.done(new SmtpAsyncClientException(FailureType.CONNECTION_FAILED_EXCEED_IDLE_MAX, sessionId, sessionCtx));
+                // closing the channel if server is not responding for max read timeout limit
+                cleanup();
```
fansu, line 118: reverse the order, close(ctx) and then cleanup()

core/src/main/java/com/yahoo/smtpnio/async/netty/PlainGreetingHandler.java
```diff
@@ -0,0 +1,154 @@
...
+        logOpt = null;
+        sessionCtx = null;
+    }
+
+    /**
+     * Closes the connection.
+     *
+     * @param ctx the ChannelHandlerContext
+     */
+    private void close(final ChannelHandlerContext ctx) {
```
fansu, line 148: @Nonnull annotation for ctx

core/src/main/java/com/yahoo/smtpnio/async/client/SslDetectHandler.java
```diff
@@ -0,0 +1,180 @@
...
+
+        // ssl failed, re-connect with plain connection
+        if (cause.getCause() instanceof NotSslRecordException) {
+
+            if (logger.isTraceEnabled() || debugOption == SmtpAsyncSession.DebugMode.DEBUG_ON) {
+                logger.debug(SSL_DETECT_REC, sessionId, sessionCtx, "Not available", host, port, true, enableStarttls, sniNames);
+            }
+
+            // if starttls is enbaled, try to create a new connection without ssl
+            if (enableStarttls) {
```
fansu, line 124: remove this flag

core/src/main/java/com/yahoo/smtpnio/async/netty/PlainGreetingHandler.java
```diff
@@ -0,0 +1,154 @@
...
+import com.yahoo.smtpnio.async.response.SmtpResponse;
+
+import io.netty.channel.Channel;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.handler.codec.MessageToMessageDecoder;
+import io.netty.handler.timeout.IdleState;
+import io.netty.handler.timeout.IdleStateEvent;
+
+/**
+ * This handler receives greeting from server after plain re-connection. If the response code if 250, start Starttls flow by sending EHLO command.
```
fansu, line 31: is 250

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
```diff
@@ -0,0 +1,169 @@
...
+
+    /** Literal for the name registered in pipeline. */
+    public static final String HANDLER_NAME = "StarttlsHandler";
```
fansu, line 41: same

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
```diff
@@ -0,0 +1,169 @@
...
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelPipeline;
+import io.netty.handler.codec.MessageToMessageDecoder;
+import io.netty.handler.ssl.SslContext;
+import io.netty.handler.ssl.SslContextBuilder;
+import io.netty.handler.timeout.IdleState;
+import io.netty.handler.timeout.IdleStateEvent;
+
+/**
+ * This handler receives STARTTLS response sent by {@link EhloHandler}. If the response code if 250, add SslHandler to upgrade current plain
```
fansu, line 35: is 250

core/src/main/java/com/yahoo/smtpnio/async/netty/EhloHandler.java
```diff
@@ -0,0 +1,152 @@
...
+     *
+     * @param sessionFuture SMTP session future
+     * @param logger logger instance
+     * @param logOpt logging option for this session.
+     * @param sessionId session id
+     * @param sessionCtx context for session information
+     * @param sessionData sessionData object containing information about the connection.
+     */
+    public EhloHandler(@Nonnull final SmtpFuture<SmtpAsyncCreateSessionResponse> sessionFuture, @Nonnull final Logger logger,
+            @Nonnull final DebugMode logOpt, final long sessionId, @Nullable final Object sessionCtx,
```
fansu, line 68: are these two variable nonnull?

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
```diff
@@ -0,0 +1,169 @@
...
+     * @param logger logger instance
+     * @param logOpt logging option for this session.
+     * @param sessionId session id
+     * @param sessionCtx context for session information
+     * @param sessionData sessionData object containing information about the connection.
+     */
+    public StarttlsHandler(@Nonnull final SmtpFuture<SmtpAsyncCreateSessionResponse> sessionFuture,
+            @Nonnull final Logger logger,
+            @Nonnull final DebugMode logOpt, final long sessionId, @Nullable final Object sessionCtx,
+            @Nullable final SmtpAsyncSessionData sessionData) {
```
fansu, line 74: are these two variable nonnull?

core/src/main/java/com/yahoo/smtpnio/async/netty/PlainGreetingHandler.java
```diff
@@ -0,0 +1,154 @@
...
+     * @param logger the {@link Logger} instance for {@link SmtpAsyncSessionImpl}
+     * @param logOpt logging option for the session to be created
+     * @param sessionId the session id
+     * @param sessionCtx context for the session information; it is used used for logging
+     * @param sessionData sessionData object containing information about the connection.
+     */
+    public PlainGreetingHandler(@Nonnull final SmtpFuture<SmtpAsyncCreateSessionResponse> sessionFuture, @Nonnull final Logger logger,
+            @Nonnull final DebugMode logOpt, final long sessionId,
+            @Nullable final Object sessionCtx,
+            @Nullable final SmtpAsyncSessionData sessionData) {
```
fansu, line 72: are these two variable nonnull?

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
```diff
@@ -0,0 +1,169 @@
...
+            try {
+                // add sslHandler
+                SslContext sslContext = SslContextBuilder.forClient().build();
+                ctx.pipeline().addFirst(SmtpAsyncClient.newSslHandler(sslContext, ctx.alloc(), sessionData.getHost(), sessionData.getPort(),
+                        sessionData.getSniNames()));
+                if (logger.isTraceEnabled() || logOpt == SmtpAsyncSession.DebugMode.DEBUG_ON) {
+                    logger.debug("[{},{}] Starttls was successful. Connection is now encrypted. ", sessionId, sessionCtx);
+                }
+                sessionCreatedFuture.done(response);
+            } catch (SSLException e) {
```
fansu, line 101: final

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
```diff
@@ -0,0 +1,169 @@
...
+        sessionCreatedFuture.done(new SmtpAsyncClientException(FailureType.CONNECTION_INACTIVE));
+        cleanup();
+    }
+
+    /**
+     * Closes the connection.
+     *
+     * @param ctx the ChannelHandlerContext
+     */
+    private void close(final ChannelHandlerContext ctx) {
```
fansu, line 151: same

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
```diff
@@ -0,0 +1,169 @@
...
+            } catch (SSLException e) {
+                logger.error("[{},{}] Failed to create SslContext after receving STARTTLS response: {}", sessionId, sessionCtx, e);
+                sessionCreatedFuture.done(new SmtpAsyncClientException(FailureType.CONNECTION_FAILED_EXCEPTION, e, sessionId, sessionCtx));
+                close(ctx);
+            }
+        } else {
+            logger.error("[{},{}] STARTTLS response was not successful:{}", sessionId, sessionCtx, serverResponse.toString());
+            sessionCreatedFuture.done(new SmtpAsyncClientException(FailureType.CONNECTION_FAILED_INVALID_GREETING_CODE, sessionId, sessionCtx,
+                    serverResponse.toString()));
+            close(ctx); // closing the channel if we r not getting a ok greeting
```
fansu, line 110: we are

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
```diff
@@ -0,0 +1,169 @@
...
+
+    @Override
+    public void userEventTriggered(@Nonnull final ChannelHandlerContext ctx, @Nonnull final Object msg) {
+        if (msg instanceof IdleStateEvent) { // Handle the IdleState if needed
+            final IdleStateEvent event = (IdleStateEvent) msg;
+            if (event.state() == IdleState.READER_IDLE) {
+                logger.error("[{},{}] Re-connection failed due to taking longer than configured allowed time.", sessionId, sessionCtx);
+                sessionCreatedFuture.done(new SmtpAsyncClientException(FailureType.CONNECTION_FAILED_EXCEED_IDLE_MAX, sessionId, sessionCtx));
+                // closing the channel if server is not responding for max read timeout limit
+                cleanup();
```
fansu, line 131: reverse the order

core/src/main/java/com/yahoo/smtpnio/async/request/StarttlsCommand.java
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * This class defines a STARTTLS (STARTTLS) command.
+ */
+public class StarttlsCommand extends AbstractSmtpCommand {
+
+    /** String literal for "STARTTLS". */
+    private static final String STARTTLS = "STARTTLS";
+
+    /**
+     * Initializes a STARTTLS command object used to log out of the server.
```
fansu, line 16: to upgrade plain connection to tls connection

core/src/main/java/com/yahoo/smtpnio/async/netty/EhloHandler.java
```diff
@@ -0,0 +1,152 @@
...
+        sessionCreatedFuture.done(new SmtpAsyncClientException(FailureType.CONNECTION_INACTIVE));
+        cleanup();
+    }
+
+    /**
+     * Closes the connection.
+     *
+     * @param ctx the ChannelHandlerContext
+     */
+    private void close(final ChannelHandlerContext ctx) {
```
fansu, line 134: nonnull

core/src/main/java/com/yahoo/smtpnio/async/client/SslDetectHandler.java
```diff
@@ -0,0 +1,180 @@
+/*
+ * Copyright Verizon Media
+ * Licensed under the terms of the Apache 2.0 license. See LICENSE file in project root for terms.
+ */
+package com.yahoo.smtpnio.async.client;
```
fansu, line 5: can you move this class to netty/ package

fansu:
please check the comments and start UT

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
cifer828: added SslHandlerBuilder.

core/src/main/java/com/yahoo/smtpnio/async/client/SslDetectHandler.java
cifer828: this class needs to create a package level classSmtpClientChannelInitializer. it can't move to another package without changing access level of SmtpClientChannelInitializer

core/src/main/java/com/yahoo/smtpnio/async/netty/EhloHandler.java
cifer828: ``sessionCtx``` can be null, ``sessionData`` should be nonnull.

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
cifer828: ```sessionCtx``` can be null, ```sessionData`` should be nonnull.

core/src/main/java/com/yahoo/smtpnio/async/netty/PlainGreetingHandler.java
cifer828: ```sessionCtx``` can be null, ```sessionData`` should be nonnull.

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncSessionData.java
cifer828: In a previous comment, nitu would like me to add this starttls flag saying if the client want to try starttls.

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
cifer828: ```sessionCtx``` can be null, ```sessionData``` should be nonnull. removed ```sessionCtx```  from the arguments because it can be obtained by ```sessionData.getSessionContext()```

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
```diff
@@ -202,9 +206,38 @@ public long applyAsLong(final long counter) { // increment by 1 unless it overfl
                 // add session specific handlers
                 switch (sessionMode) {
                 case SSL:
-                    // add sslHandler for initial ssl connection
+                    // create sslHandler for ssl connection
                     final SslHandler sslHandler = SslHandlerBuilder.newBuilder(sslContext, ch.alloc(), host, port, sniNames).build();
-                    pipeline.addFirst(SSL_HANDLER, sslHandler);
+                    Promise<Channel> sslConnectionFture = (Promise<Channel>) sslHandler.handshakeFuture();
+                    // check if ssl connection succeed
+                    sslConnectionFture.addListener(new GenericFutureListener<io.netty.util.concurrent.Future<? super Channel>>() {
```
cifer828, line 213: I tried to do re-connection here but failed. The reason is that ```exceptionCaught``` will still be triggered after  operationComplete() has been called. So we still need a customized exceptionCaught() to stop the "exceptionCaught chain" reaching ```SmtpConnectionHandler```. Because if it reaches there, ```sessionFuture``` will be done with exception which we don't want for starttls flow. So I think it's better to just do reconnection in the customized exceptionCaught().

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
```diff
@@ -202,9 +206,38 @@ public long applyAsLong(final long counter) { // increment by 1 unless it overfl
...
+                        @Override
+                        public void operationComplete(io.netty.util.concurrent.Future<? super Channel> future) throws Exception {
+                            if (!future.isSuccess()) {
+                                ch.close();
+                                // fail early if ssl connection can't established
+                                try {
+                                    future.get();
```
cifer828, line 220: to get the exception.

core/src/main/java/com/yahoo/smtpnio/async/client/SslDetectHandler.java
```diff
@@ -166,6 +165,13 @@ private void close(final ChannelHandlerContext ctx) {
         }
     }
 
+    @Override
+    public void channelInactive(@Nonnull final ChannelHandlerContext ctx) {
```
cifer828, line 169: I add this method to stop the "channelInactive chain" reaching ```SmtpConnectionHandler``` after calling ```channel.close()``` before re-connection. If we remove this method, ```channelInactive()``` in ```SmtpConnectionHandler``` will fail the sessionFuture with exception. 
But if we add ```SmtpConnectionHandler```  in the ```decode()```, we don't need this method.

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
```diff
@@ -68,6 +87,9 @@
     /** Event loop group that will serve all channels for the SMTP client. */
     private final EventLoopGroup group;
 
+    /** This SmtpAsyncClient object . */
+    private final SmtpAsyncClient smtpAsyncClient;
```
fansu, line 91: do not store it as member variable.

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
fansu: please add timeout value and time unit here.

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
```diff
@@ -132,47 +183,81 @@ public void operationComplete(final io.netty.util.concurrent.Future<? super Void
...
+                    sslConnectionFture.addListener(new GenericFutureListener<io.netty.util.concurrent.Future<? super Channel>>() {
+                        @Override
+                        public void operationComplete(io.netty.util.concurrent.Future<? super Channel> future) throws Exception {
+                            if (!future.isSuccess()) {
+                                ch.close();
+                                // fail early if ssl connection can't established
+                                try {
+                                    future.get();
+                                } catch (ExecutionException e) {
+                                    if (e.getCause() instanceof NotSslRecordException && isStarttls) {
```
fansu, line 222: i do not understand, why do you need to check for NotSslRecordException? if not, remove.

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
fansu: actually, you do not need to do this. you can create SmtpAsyncClientException with cause. This is already a failure case

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
```diff
@@ -132,47 +183,81 @@ public void operationComplete(final io.netty.util.concurrent.Future<? super Void
...
+                                SmtpAsyncClientException ex = new SmtpAsyncClientException(
+                                        SmtpAsyncClientException.FailureType.CONNECTION_FAILED_EXCEPTION, future.cause());
+                                logger.error(CONNECT_RESULT_REC, sessionId, sessionCtx, "failure", host, port, isSsl, sniNames, ex);
+                                sessionCreatedFuture
+                                        .done(new SmtpAsyncClientException(FailureType.CONNECTION_FAILED_EXCEPTION, sessionId, sessionCtx));
+                            }
+                        }
+                    });
+                    // pipeline: IdleHandler, SslHandler, SslDetectHandler, other initializer handlers, SmtpClientConnectHandler
+                    pipeline.addAfter(IDLE_STATE_HANDLER_NAME, SslDetectHandler.HANDLER_NAME, new SslDetectHandler(sessionCount.get(), sessionData,
```
fansu, line 236: this block is super confusing with addAfter and addBefore. Can u somehow create a list and build it in a nicer way.

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
```diff
@@ -132,47 +183,81 @@ public void operationComplete(final io.netty.util.concurrent.Future<? super Void
...
+                final SessionMode sessionMode = isSsl ? SessionMode.SSL : (isStarttls ? SessionMode.STARTTLS : SessionMode.NON_SSL);
+
+                // add session specific handlers
+                switch (sessionMode) {
+                case SSL:
+                    // create sslHandler for ssl connection
+                    final SslHandler sslHandler = SslHandlerBuilder.newBuilder(sslContext, ch.alloc(), host, port, sniNames).build();
+                    Promise<Channel> sslConnectionFture = (Promise<Channel>) sslHandler.handshakeFuture();
+                    // check if ssl connection succeed
+                    sslConnectionFture.addListener(new GenericFutureListener<io.netty.util.concurrent.Future<? super Channel>>() {
```
fansu, line 213: this a chain of future and listener, can you double check and see if you can make it one layer. This will be super hard for debugging in the future.

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
cifer828: use two addAfter(IDLE_STATE_HANDLER_NAME)

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
cifer828: removed.

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
cifer828: change it to a local variable

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
cifer828: remove this listener.

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
cifer828: removed this listener because if sessionFuture will be done with exception twice. once here, once in the exceptionCaught method in handlers.

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
cifer828: fixed.

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
```diff
@@ -132,47 +176,57 @@ public void operationComplete(final io.netty.util.concurrent.Future<? super Void
...
+                // decide session state
+                final SessionMode sessionMode = isSsl ? SessionMode.SSL : (isStarttls ? SessionMode.STARTTLS : SessionMode.NON_SSL);
+
+                // add session specific handlers
+                switch (sessionMode) {
+                case SSL:
+                    // create sslHandler for ssl connection
+                    final SslHandler sslHandler = SslHandlerBuilder.newBuilder(sslContext, ch.alloc(), host, port, sniNames).build();
+                    // pipeline: IdleHandler, SslHandler, SslDetectHandler, other initializer handlers, SmtpClientConnectHandler
+                    pipeline.addAfter(IDLE_STATE_HANDLER_NAME, SslDetectHandler.HANDLER_NAME, new SslDetectHandler(sessionCount.get(), sessionData,
```
nituyahoo, line 205: can we add ssl detect handler only if the client specifies in the config to support starttls

core/src/main/java/com/yahoo/smtpnio/async/client/SslDetectHandler.java
```diff
@@ -0,0 +1,184 @@
...
+                // if starttls is not enabled, finish the future with exception
+                final SmtpAsyncClientException ex = new SmtpAsyncClientException(SmtpAsyncClientException.FailureType.NOT_SSL_RECORD,
+                        cause.getCause());
+                logger.error(SmtpAsyncClient.CONNECT_RESULT_REC, sessionId, sessionCtx, "failure", host, port, false, sniNames, ex);
+                sessionCreatedFuture.done(ex);
+                cleanup();
+            }
+        } else {
+            logger.error("[{},{}] Connection failed due to encountering exception:{}.", sessionId, sessionCtx, cause);
+            sessionCreatedFuture.done(new SmtpAsyncClientException(FailureType.CONNECTION_FAILED_EXCEPTION, cause, sessionId, sessionCtx));
```
nituyahoo, line 132: why are we completing the future here? what happens in case where the exception gets throws for ssl connection and the start tls option is not given? where does the future get completed? please discuss

core/src/main/java/com/yahoo/smtpnio/async/client/SslDetectHandler.java
```diff
@@ -0,0 +1,184 @@
...
+
+        }
+        cleanup();
+    }
+
+    @Override
+    public void userEventTriggered(@Nonnull final ChannelHandlerContext ctx, @Nonnull final Object msg) {
+        if (msg instanceof IdleStateEvent) { // Handle the IdleState if needed
+            final IdleStateEvent event = (IdleStateEvent) msg;
+            if (event.state() == IdleState.READER_IDLE) {
```
nituyahoo, line 142: same question as above where is idle event normally handled?

core/src/main/java/com/yahoo/smtpnio/async/client/SslDetectHandler.java
nituyahoo: please discuss

core/src/main/java/com/yahoo/smtpnio/async/netty/PlainReconnectGreetingHandler.java
```diff
@@ -0,0 +1,151 @@
...
+    public void exceptionCaught(@Nonnull final ChannelHandlerContext ctx, @Nonnull final Throwable cause) {
+        logger.error("[{},{}] Re-connection failed due to encountering exception:{}.", sessionId, sessionCtx, cause);
+        sessionCreatedFuture.done(new SmtpAsyncClientException(FailureType.CONNECTION_FAILED_EXCEPTION, cause, sessionId, sessionCtx));
+        close(ctx);
+        cleanup();
+    }
+
+    @Override
+    public void userEventTriggered(@Nonnull final ChannelHandlerContext ctx, @Nonnull final Object msg) {
+        if (msg instanceof IdleStateEvent) { // Handle the IdleState if needed
```
nituyahoo, line 108: similar comment as the above handler

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsEhloHandler.java
```diff
@@ -0,0 +1,171 @@
...
+    public void exceptionCaught(@Nonnull final ChannelHandlerContext ctx, @Nonnull final Throwable cause) {
+        logger.error("[{},{}] Re-connection failed due to encountering exception:{}.", sessionId, sessionCtx, cause);
+        sessionCreatedFuture.done(new SmtpAsyncClientException(FailureType.CONNECTION_FAILED_EXCEPTION, cause, sessionId, sessionCtx));
+        close(ctx);
+        cleanup();
+    }
+
+    @Override
+    public void userEventTriggered(@Nonnull final ChannelHandlerContext ctx, @Nonnull final Object msg) {
+        if (msg instanceof IdleStateEvent) { // Handle the IdleState if needed
```
nituyahoo, line 126: similar comment as above handler

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
cifer828: yes we can. may i ask the reason for that? is it because we want to use as less handlers as possible?

core/src/main/java/com/yahoo/smtpnio/async/client/SslDetectHandler.java
cifer828: close() can be kept to close the channel when expected exception is caught

core/src/main/java/com/yahoo/smtpnio/async/client/SslDetectHandler.java
cifer828: i think if we get other exceptions instead of NotSslRecordException, it's not part of the starttls flow, so we should just fail here.  when NotSslRecordException is thrown but starttls is disabled, this method also fails the future with exception in the above branch.

core/src/main/java/com/yahoo/smtpnio/async/netty/PlainReconnectGreetingHandler.java
cifer828: in the starttls flow, we don't have SmtpClientConnectHandler, so the idle event will be handled by the three dynamic handlers PlainReconnectGreetingHandler, StarttlsEhloHandler and StarttlsSessionHandler.

core/src/main/java/com/yahoo/smtpnio/async/client/SslDetectHandler.java
cifer828: idle event will be handled in the first handler with userEventTriggered() after IdleStateHandler. in the latest commit, handler order in pipeline is SslHandler, SslDetectHandler, IdleStateHandler .... SmtpClientConnectionHandler. so the idle event will be handled in SmtpClientConnectionHandler. i'll remove this method here.

cifer828:
use one handler version instead

PR #14, opened 2020-07-24
Use one handler to implement starttls flow.

cifer828:
## Description

Current flow
- If ```enableSsl``` is true, add ```SslHandler``` and ```SslDetectHandler``` to pipeline.
- If ```SslDetectHandler``` works fine, remove it and the pipeline works as normal. If ```SslDetectHandler``` catches```NotSslRecordException``` and starttls is enabled according to session config, close current channel and create a new one without ssl. Add ```StarttlsHandler``` to start the Starttls flow.
- Starttls flow contains three steps
  1. process server connection greeting and send EHLO.
  2. process EHLO response and send STARTTLS.
  3. process STARTLS response and upgrade plain connection to ssl connection.

## Motivation and Context

STARTTLS is required for Outlook connection, since Outlook smtp server doesn't support general SSL/TLS connection. 
This is a 1 starttls handler version. 3 starttls handlers version is [here](https://github.com/yahoo/smtp-client-nio/pull/13). 

## How Has This Been Tested?

Prototype for review, no UTs added.

## Types of changes

## Checklist:

- [ ] All new and existing tests passed.


core/src/main/java/com/yahoo/smtpnio/async/client/SslDetectHandler.java
```diff
@@ -0,0 +1,175 @@
...
+     */
+    private void close(final ChannelHandlerContext ctx) {
+        if (ctx.channel().isActive()) {
+            // closing the channel if server is still active
+            ctx.close();
+        }
+    }
+
+    @Override
+    public void channelInactive(@Nonnull final ChannelHandlerContext ctx) {
```
cifer828, line 157: I add this method to stop the "channelInactive chain" reaching SmtpConnectionHandler after actively calling channel.close(). Without this method, channelInactive()will keep fired to SmtpConnectionHandler and fail the sessionFuture with exception and cut the starttls flow.

But since we don't fail here, I am afraid it will omit some unexpected channelInactive cases and cause problems. Another workaround is to dynamically add SmtpConnectionHandler in the handler'sdecode() instead of in the session creation.

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
```diff
@@ -132,47 +173,56 @@ public void operationComplete(final io.netty.util.concurrent.Future<? super Void
...
+                final SessionMode sessionMode = isSsl ? SessionMode.SSL : (isStarttls ? SessionMode.STARTTLS : SessionMode.NON_SSL);
+
+                // add session specific handlers
+                switch (sessionMode) {
+                case SSL:
+                    // handlers order in pipeline: IdleHandler, SslHandler, SslDetectHandler, other initializer handlers, SmtpClientConnectHandler
+                    pipeline.addAfter(IDLE_STATE_HANDLER_NAME, SslDetectHandler.HANDLER_NAME, new SslDetectHandler(sessionCount.get(), sessionData,
+                            config, LoggerFactory.getLogger(SslDetectHandler.class), debugOption, smtpAsyncClient, sessionCreatedFuture));
+                    final SslHandler sslHandler = SslHandlerBuilder.newBuilder(sslContext, ch.alloc(), host, port, sniNames).build();
+                    pipeline.addAfter(IDLE_STATE_HANDLER_NAME, SslHandlerBuilder.SSL_HANDLER, sslHandler);
```
cifer828, line 203: Didn't check ssl connection for the normal flow. Because if is Ssl connection fails, exceptionCaught chain will be triggered and reaches ```SmtpClientConnectionHandler```. If we also fail here, we may set session future done with exception twice (once here, once in ```SmtpClientConnectionHandler```)

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
```diff
@@ -0,0 +1,279 @@
...
+ * <li>1. process server connection greeting and send EHLO.
+ *
+ * <li>2. process EHLO response and send STARTTLS.
+ *
+ * <li>3. process STARTLS response and upgrade plain connection to ssl connection.
+ */
+public class StarttlsHandler extends MessageToMessageDecoder<SmtpResponse> {
+
+    /** This enum indicates different states in the starttls flow. */
+    private enum StarttlsState {
```
jui8wang, line 53: StartTlsState

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
```diff
@@ -0,0 +1,279 @@
...
+        /** Receives greeting after re-connection, handler will send EHLO. */
+        PRE_EHLO,
+
+        /** EHLO has been sent, handler is checking server responses and will send STARTTLS. */
+        PRE_STARTTLS,
+
+        /** STARTTLS has been sent, handler will remove itself and add SslHandler. */
```
jui8wang, line 60: Lets mention in java doc 2 things: 1. what it is waiting for,  2.  what it will do when good and bad.

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
```diff
@@ -0,0 +1,279 @@
...
+                starttlsState = StarttlsState.STARTTLS_FAILURE;
+            }
+            break;
+
+        default:
+            break;
+        }
+
+        // close channel and clean up if starttls failed
+        if (starttlsState == StarttlsState.STARTTLS_FAILURE) {
```
jui8wang, line 218: this should be part of the switch statement

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
```diff
@@ -0,0 +1,279 @@
...
+                final SmtpAsyncCreateSessionResponse response = new SmtpAsyncCreateSessionResponse(session, serverResponse);
+                try {
+                    // create sslHandler
+                    final SslContext sslContext = SslContextBuilder.forClient().build();
+                    final SslHandler sslHandler = SslHandlerBuilder
+                            .newBuilder(sslContext, ctx.alloc(), sessionData.getHost(), sessionData.getPort(), sessionData.getSniNames()).build();
+                    final Promise<Channel> sslConnectionFture = (Promise<Channel>) sslHandler.handshakeFuture();
+                    final SmtpFuture<SmtpAsyncCreateSessionResponse> sesssionFuture = this.sessionCreatedFuture;
+                    // check if ssl connection succeeds
+                    sslConnectionFture.addListener(new GenericFutureListener<io.netty.util.concurrent.Future<? super Channel>>() {
```
jui8wang, line 187: could u create an inner class called SSLHandShakeCompleteListener instead of using anonymous one.  So we can have a clear code flow just for this state.

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
```diff
@@ -0,0 +1,279 @@
...
+            if (serverResponse.getCode().value() == SmtpResponse.Code.GREETING) { // successful response
+                // add the command response handler
+                final SmtpAsyncSessionImpl session = new SmtpAsyncSessionImpl(ctx.channel(), logger, logOpt, sessionId, pipeline, sessionCtx);
+                final SmtpAsyncCreateSessionResponse response = new SmtpAsyncCreateSessionResponse(session, serverResponse);
+                try {
+                    // create sslHandler
+                    final SslContext sslContext = SslContextBuilder.forClient().build();
+                    final SslHandler sslHandler = SslHandlerBuilder
+                            .newBuilder(sslContext, ctx.alloc(), sessionData.getHost(), sessionData.getPort(), sessionData.getSniNames()).build();
+                    final Promise<Channel> sslConnectionFture = (Promise<Channel>) sslHandler.handshakeFuture();
```
jui8wang, line 184: rename it to sslHandshakeFuture

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
```diff
@@ -0,0 +1,279 @@
...
+            close(ctx);
+            cleanup();
+        }
+
+    }
+
+    @Override
+    public void exceptionCaught(@Nonnull final ChannelHandlerContext ctx, @Nonnull final Throwable cause) {
+        logger.error("[{},{}] Re-connection failed due to encountering exception:{}.", sessionId, sessionCtx, cause);
+        sessionCreatedFuture.done(new SmtpAsyncClientException(FailureType.CONNECTION_FAILED_EXCEPTION, cause, sessionId, sessionCtx));
```
jui8wang, line 229: Have a private method called handleSessionFailed() and it does the following and try to call this handleSessionFailed() methods as much as possible, so we dont miss those very important steps:
sessionCreatedFuture.done(..exception)
close(ctx);
cleanup()

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
```diff
@@ -0,0 +1,279 @@
...
+
+    @Override
+    public void decode(@Nonnull final ChannelHandlerContext ctx, @Nonnull final SmtpResponse serverResponse, @Nonnull final List<Object> out) {
+        final Channel channel = ctx.channel();
+        final ChannelPipeline pipeline = ctx.pipeline();
+
+        // all server responses should be 2xx
+        if (serverResponse.getReplyType() != SmtpResponse.ReplyType.POSITIVE_COMPLETION) {
+            // receive a bad response,
+            logger.error("[{},{}] Receive bad response from server: {}", sessionId, sessionCtx, serverResponse.toString());
```
jui8wang, line 132: remove all your logger.error.  Create an exception so that u can set it sesssionFuture.done().  In STARTTLS_FAILURE state, u do 1 logger.error

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
```diff
@@ -0,0 +1,279 @@
...
+            starttlsState = StarttlsState.STARTTLS_FAILURE;
+        }
+
+        switch (starttlsState) {
+        case PRE_EHLO:
+            // receive greeting and send EHLO
+            if (serverResponse.getCode().value() == SmtpResponse.Code.GREETING) {
+                channel.writeAndFlush(new ExtendedHelloCommand(EHLO_CLIENT_NAME).getCommandLineBytes());
```
jui8wang, line 140: what is the difference between ehlo and hlo? can we always send ehlo command?  Check SMTP sun lib

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
```diff
@@ -0,0 +1,279 @@
...
+ * <li>2. process EHLO response and send STARTTLS.
+ *
+ * <li>3. process STARTLS response and upgrade plain connection to ssl connection.
+ */
+public class StarttlsHandler extends MessageToMessageDecoder<SmtpResponse> {
+
+    /** This enum indicates different states in the starttls flow. */
+    private enum StarttlsState {
+        /** Receives greeting after re-connection, handler will send EHLO. */
+        PRE_EHLO,
```
jui8wang, line 55: rename it to GET_SERVER_GREETING

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
```diff
@@ -0,0 +1,279 @@
...
+
+        /** EHLO has been sent, handler is checking server responses and will send STARTTLS. */
+        PRE_STARTTLS,
```
jui8wang, line 58: rename it to GET_EHLO_RESP

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
```diff
@@ -0,0 +1,279 @@
...
+
+        /** STARTTLS has been sent, handler will remove itself and add SslHandler. */
+        POST_STARTTLS,
```
jui8wang, line 61: rename it to GET_STARTTLS_RESP

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
```diff
@@ -0,0 +1,279 @@
...
+                        serverResponse.toString());
+                starttlsState = StarttlsState.STARTTLS_FAILURE;
+            }
+            break;
+
+        case PRE_STARTTLS:
+            // receive EHLO responses and send STARTTLS
+            final String message = serverResponse.getMessage();
+            // check STARTTLS capability
+            if (message.length() >= STARTTLS.length() && message.substring(0, STARTTLS.length()).equalsIgnoreCase(STARTTLS)) {
```
jui8wang, line 153: we dont have a parser class to parse those?

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
```diff
@@ -0,0 +1,279 @@
...
+                receivedStarttlsCapability = true;
+            }
+
+            // decide next state when we receive last response
+            if (serverResponse.isLastLineResponse()) {
+                if (receivedStarttlsCapability) {
+                    channel.writeAndFlush(new StarttlsCommand().getCommandLineBytes());
+                    starttlsState = StarttlsState.POST_STARTTLS;
```
jui8wang, line 161: can u put this. for all member variables

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
cifer828: remove this variable by directly calling sslHandler.handshakeFuture().addListener().

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
cifer828: In SMTPTransport. protocolConnect(), it first check if EHLO succeed with code 250, if not, it will send HELO as fallback. I add this logic into our code.

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
cifer828: only add this to member variables in this class and SslDetectHandler. didn't change other existing classes.

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
cifer828: fixed by comparing message with "STARTTS\r\n"

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
cifer828: removed this branch. use a boolean variable instead because failure is not a state.

core/src/main/java/com/yahoo/smtpnio/async/netty/SslDetectHandler.java
```diff
@@ -102,48 +98,16 @@ public void exceptionCaught(@Nonnull final ChannelHandlerContext ctx, @Nonnull f
...
-            if (enableStarttls) {
-                this.smtpAsyncClient.createStarttlsSession(this.sessionData, this.sessionConfig, this.logOpt, this.sessionCreatedFuture);
-            } else {
-                // if startTls is not enabled, finish the future with exception
-                final SmtpAsyncClientException ex = new SmtpAsyncClientException(SmtpAsyncClientException.FailureType.NOT_SSL_RECORD,
-                        cause.getCause());
-                this.logger.error(SmtpAsyncClient.CONNECT_RESULT_REC, this.sessionId, sessionCtx, "failure", host, port, false, sniNames, ex);
-                this.sessionCreatedFuture.done(ex);
-            }
-        } else {
```
cifer828: To reflect this comment https://github.com/yahoo/smtp-client-nio/pull/13#discussion_r461175031. SslDetectHandler only handle NotSslRecordException for startTls. other exceptions will be handled by SmtpClientConnectionHandler.

core/src/main/java/com/yahoo/smtpnio/async/netty/SslDetectHandler.java
```diff
@@ -102,48 +98,16 @@ public void exceptionCaught(@Nonnull final ChannelHandlerContext ctx, @Nonnull f
...
-            this.logger.error("[{},{}] Connection failed due to encountering exception:{}.", this.sessionId, sessionCtx, cause);
-            this.sessionCreatedFuture.done(new SmtpAsyncClientException(FailureType.CONNECTION_FAILED_EXCEPTION, cause, this.sessionId, sessionCtx));
-
-        }
-        cleanup();
-    }
-
-    @Override
-    public void userEventTriggered(@Nonnull final ChannelHandlerContext ctx, @Nonnull final Object msg) {
```
cifer828: similar as above.

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
```diff
@@ -0,0 +1,321 @@
...
+                final SmtpAsyncCreateSessionResponse createSessionResponse = new SmtpAsyncCreateSessionResponse(session, serverResponse);
+                try {
+                    // create sslHandler
+                    final SslContext sslContext = SslContextBuilder.forClient().build();
+                    final SslHandler sslHandler = SslHandlerBuilder.newBuilder(sslContext, ctx.alloc(), this.sessionData.getHost(),
+                            this.sessionData.getPort(), this.sessionData.getSniNames()).build();
+                    // add listener to check if ssl connection succeeds
+                    sslHandler.handshakeFuture().addListener(new SSLHandShakeCompleteListener(createSessionResponse, ctx));
+                    ctx.pipeline().addFirst(SslHandlerBuilder.SSL_HANDLER, sslHandler);
+                } catch (final SSLException e) {
```
cifer828, line 245: this exception is thrown by line 239 which uses static method to create a SslContext. sure how to cover this in UT.

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
```diff
@@ -104,75 +117,133 @@ public SmtpAsyncClient(final int numThreads) throws SSLException {
...
+                            LoggerFactory.getLogger(StarttlsHandler.class), debugOption, sessionId, sessionData));
+                    break;
+
+                case NON_SSL:
+                    // add SmtpClientConnectHandler for initial non-ssl connection
+                    pipeline.addLast(SmtpClientConnectHandler.HANDLER_NAME, new SmtpClientConnectHandler(sessionCreatedFuture,
+                            LoggerFactory.getLogger(SmtpClientConnectHandler.class), debugOption, sessionId, sessionCtx));
+                    break;
+
+                default:
```
cifer828, line 237: not sure how to cover default branch in UT

core/src/main/java/com/yahoo/smtpnio/async/netty/SslDetectHandler.java
```diff
@@ -0,0 +1,145 @@
...
+
+        final String host = this.sessionData.getHost();
+        final int port = this.sessionData.getPort();
+        final Collection<String> sniNames = this.sessionData.getSniNames();
+        final Object sessionCtx = this.sessionData.getSessionContext();
+
+        // ssl failed, re-connect with plain connection
+        if (cause.getCause() instanceof NotSslRecordException) {
+            close(ctx);
+            // if startTls is enbaled, try to create a new connection without ssl
```
fansu, line 105: typo, enabled

core/src/main/java/com/yahoo/smtpnio/async/netty/SslDetectHandler.java
```diff
@@ -0,0 +1,145 @@
...
+    private void close(final ChannelHandlerContext ctx) {
+        if (ctx.channel().isActive()) {
+            // closing the channel if server is still active
+            ctx.close();
+        }
+    }
+
+    @Override
+    public void channelInactive(@Nonnull final ChannelHandlerContext ctx) {
+        // after close() has been called on SslHandler failure, channelInactive() will also be called
```
fansu, line 128: SslDetectHandler

core/src/main/java/com/yahoo/smtpnio/async/netty/SslHandlerBuilder.java
```diff
@@ -0,0 +1,105 @@
...
+
+    /** Port number to connect to. */
+    private final int port;
+
+    /** Server Name Indication (SNI) list. */
+    @Nullable
+    private final Collection<String> sniNames;
+
+    /**
+     * Initialize an SslHandlerBuilder object used to build {@link SslHanlder}.
```
fansu, line 50: typo SslHandler

core/src/main/java/com/yahoo/smtpnio/async/netty/SslHandlerBuilder.java
```diff
@@ -0,0 +1,105 @@
...
+            @Nullable final Collection<String> sniNames) {
+        this.sslContext = sslContext;
+        this.allocator = alloc;
+        this.host = host;
+        this.port = port;
+        this.sniNames = sniNames;
+    }
+
+    /**
+     * Creates a new SslHandlerBuilder object used to build {@link SslHanlder}.
```
fansu, line 68: same

core/src/main/java/com/yahoo/smtpnio/async/netty/SslHandlerBuilder.java
```diff
@@ -0,0 +1,105 @@
...
+     * @param sniNames collection of SNI names
+     * @return a SslHandlerBuilder object used to build {@link SslHanlder}
+     */
+    public static SslHandlerBuilder newBuilder(@Nonnull final SslContext sslContext, @Nonnull final ByteBufAllocator alloc,
+            @Nonnull final String host, final int port, @Nullable final Collection<String> sniNames) {
+        return new SslHandlerBuilder(sslContext, alloc, host, port, sniNames);
+    }
+
+    /**
+     * Create a new {@link SslHanlder} for decryption/encryption.
```
fansu, line 83: same

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
```diff
@@ -0,0 +1,321 @@
...
+                this.startTlsState = StartTlsState.GET_EHLO_RESP;
+            } else {
+                this.isSessionFutureFailed = true;
+            }
+            break;
+
+        case GET_EHLO_RESP:
+            // check EHLO responses
+            if (serverResponse.getMessage().equalsIgnoreCase(STARTTLS)) {
+                this.receivedStarttlsCapability = true;
```
fansu, line 189: I saw this flag is switch back to false in line 197. Can you only set this flag 1 time instead of 2? It will be buggy in the future

fansu: combine this block with following blocks

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
```diff
@@ -0,0 +1,321 @@
...
+                } else {
+                    // server doesn't reply STARTTLS capability
+                    this.isSessionFutureFailed = true;
+                }
+            }
+            break;
+
+        case GET_HELO_RESP:
+            // check HELO responses
+            if (serverResponse.getMessage().equalsIgnoreCase(STARTTLS)) {
```
fansu, line 212: same

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
fansu: Also, do you need receivedStarttlsCapability to be class variable? Local variable should be ok, I do not see you use it anywhere else and it is safer to use local variable

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
```diff
@@ -0,0 +1,321 @@
...
+            break;
+
+        default:
+            // shouldn't reach here
+            ctx.fireExceptionCaught(new SmtpAsyncClientException(FailureType.STARTTLS_FALIED, "Invalid state"));
+        }
+
+        // close channel and clean up if startTls failed
+        if (this.isSessionFutureFailed) {
+            this.logger.error("[{},{}] startTls failed, server response: {}", sessionId, sessionCtx, serverResponse.toString());
```
fansu, line 260: you lost all exceptions above

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
```diff
@@ -0,0 +1,321 @@
...
+                try {
+                    // create sslHandler
+                    final SslContext sslContext = SslContextBuilder.forClient().build();
+                    final SslHandler sslHandler = SslHandlerBuilder.newBuilder(sslContext, ctx.alloc(), this.sessionData.getHost(),
+                            this.sessionData.getPort(), this.sessionData.getSniNames()).build();
+                    // add listener to check if ssl connection succeeds
+                    sslHandler.handshakeFuture().addListener(new SSLHandShakeCompleteListener(createSessionResponse, ctx));
+                    ctx.pipeline().addFirst(SslHandlerBuilder.SSL_HANDLER, sslHandler);
+                } catch (final SSLException e) {
+                    this.isSessionFutureFailed = true;
```
fansu, line 246: you lost exception here, for logger and handleSessionFailed

core/src/main/java/com/yahoo/smtpnio/async/netty/SslDetectHandler.java
cifer828: this comments means that when SslHandler fails, this channelInactive() will be called since we close the channel in exceptionCaught()

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
cifer828: yes, it needs to be a member variable since the decode() method will be called several time when receiving EHLO response because the previous SmtpClientRespReader will split server response based on "\r\n"

cifer828: > I saw this flag is switch back to false in line 197. Can you only set this flag 1 time instead of 2? It will be buggy in the future

this flag is set to false here because if EHLO fails, we try HELO as a fallback and all the state should be restored to origin. but i think the server won't reply STARTTLS with a bad response code, so if EHLO fails, we shouldn't receive STARTTLS. so i just remove this line.

cifer828: > combine this block with following blocks

this decode() method will be called several times in the GET_EHLO_RESP state. we check STARTTLS every time when decode() is called, but only decide next state when receiving the last response. so i make two condition block.

core/src/test/java/com/yahoo/smtpnio/async/netty/SslDetectHandlerTest.java
```diff
@@ -0,0 +1,429 @@
...
+    /**
+     * Tests {@code exceptionCaught} method when NotSslRecordException is caught with logger trace off but debug on.
+     *
+     * @throws IllegalArgumentException will not throw in this test
+     * @throws InterruptedException will not throw in this test
+     * @throws TimeoutException will not throw in this test
+     * @throws IllegalAccessException will not throw in this test
+     */
+    @Test
+    public void testtNotSslRecordExceptionCaughtDebugOn()
```
fansu, line 240: typo, test*

core/src/test/java/com/yahoo/smtpnio/async/netty/StarttlsHandlerTest.java
```diff
@@ -0,0 +1,712 @@
+package com.yahoo.smtpnio.async.netty;
```
fansu, line 1: add licence info here

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
fansu: as discussed, lets not make change here.

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
fansu: as discussed, lets not make change here.

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
```diff
@@ -0,0 +1,317 @@
...
+            }
+            break;
+
+        case GET_HELO_RESP:
+            // check HELO responses
+            if (serverResponse.getMessage().equalsIgnoreCase(STARTTLS)) {
+                this.receivedStarttlsCapability = true;
+            }
+
+            if (serverResponse.isLastLineResponse()) {
```
fansu, line 212: Check for HELO not success as well

cifer828: sure

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
cifer828: as discussed, no change is made here

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
cifer828: sure

core/src/main/java/com/yahoo/smtpnio/async/netty/SslDetectHandler.java
```diff
@@ -0,0 +1,145 @@
...
+    @Override
+    public void exceptionCaught(@Nonnull final ChannelHandlerContext ctx, @Nonnull final Throwable cause) {
+
+        final String host = this.sessionData.getHost();
+        final int port = this.sessionData.getPort();
+        final Collection<String> sniNames = this.sessionData.getSniNames();
+        final Object sessionCtx = this.sessionData.getSessionContext();
+
+        // ssl failed, re-connect with plain connection
+        if (cause.getCause() instanceof NotSslRecordException) {
```
jui8wang, line 103: what happens to other cause?

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
```diff
@@ -104,75 +117,133 @@ public SmtpAsyncClient(final int numThreads) throws SSLException {
...
 
+    /**
+     * Re-connect to the remote server using plain connection and start the Starttls flow.
+     *
+     * @param sessionData a {@link SmtpAsyncSessionData} object containing information about the connection
+     * @param config configuration to be used for this session/connection
+     * @param debugOption the debugging option used
+     * @param sessionCreatedFuture future containing the result of the request
+     */
+    public void createStarttlsSession(@Nonnull final SmtpAsyncSessionData sessionData, @Nonnull final SmtpAsyncSessionConfig config,
```
jui8wang, line 135: no public, do u plan to let the caller uses it or for internal ? If it is only for internal, u should make it private level and move the SslDetectHandler as an inner class.   Lets discuss.

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
```diff
@@ -0,0 +1,319 @@
...
+         */
+        public SSLHandShakeCompleteListener(@Nonnull final SmtpAsyncCreateSessionResponse createSessionResponse,
+                @Nonnull final ChannelHandlerContext ctx) {
+            this.createSessionResponse = createSessionResponse;
+            this.ctx = ctx;
+        }
+
+        @Override
+        public void operationComplete(@Nonnull final Future<? super Channel> sslConnectionFuture) throws Exception {
+            if (sslConnectionFuture.isSuccess()) {
```
jui8wang, line 82: why is it called sslConnectionFuture, i assume all it does is hand shake

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
```diff
@@ -0,0 +1,319 @@
...
+        if (isSessionFutureFailed) {
+            this.logger.error("[{},{}] startTls failed, server response: {}", this.sessionId, this.sessionCtx, serverResponse.toString());
+            handleSessionFailed(ctx,
+                    new SmtpAsyncClientException(FailureType.STARTTLS_FALIED, this.sessionId, this.sessionCtx, serverResponse.toString()));
+        }
+    }
+
+    @Override
+    public void exceptionCaught(@Nonnull final ChannelHandlerContext ctx, @Nonnull final Throwable cause) {
+        this.logger.error("[{},{}] Starttls Connection failed due to encountering exception:{}.", this.sessionId, this.sessionCtx, cause);
```
jui8wang, line 265: change the logging to debug instead of error

jui8wang: Please minimize the frequency in logging with error.  we can discuss.

core/src/main/java/com/yahoo/smtpnio/async/netty/SslHandlerBuilder.java
```diff
@@ -0,0 +1,105 @@
...
+import javax.net.ssl.SSLParameters;
+
+import io.netty.buffer.ByteBufAllocator;
+import io.netty.handler.ssl.SslContext;
+import io.netty.handler.ssl.SslHandler;
+
+/**
+ * This class provides a builder to create new SslHandler.
+ */
+public final class SslHandlerBuilder {
```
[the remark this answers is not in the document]
cifer828: if i create a build() method in SmtpAsyncClient, it needs to be public static because StarttlsHandler will also call it. is that OK

core/src/main/java/com/yahoo/smtpnio/async/netty/SslDetectHandler.java
cifer828: missed ctx.fireExceptionCaught(cause) in else statement, will add.

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
cifer828: SslDetectHandler will only be created in SmtpAsyncClient, i think we can make it as an inner class so that createStarttlsSession() can be private.

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
cifer828: this future record the result of TLS handshake, so i think this future's success can indicate the ssl connection success.  maybe sslHandshakeFuture is a better name?

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
cifer828: i think we still need to log error here because the session future failed here. if we change this logger to debug level, then we should also change all other the logger.error in this class to ensure the consistency.

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
```diff
@@ -0,0 +1,319 @@
...
+        public SSLHandShakeCompleteListener(@Nonnull final SmtpAsyncCreateSessionResponse createSessionResponse,
+                @Nonnull final ChannelHandlerContext ctx) {
+            this.createSessionResponse = createSessionResponse;
+            this.ctx = ctx;
+        }
+
+        @Override
+        public void operationComplete(@Nonnull final Future<? super Channel> sslConnectionFuture) throws Exception {
+            if (sslConnectionFuture.isSuccess()) {
+                sessionCreatedFuture.done(this.createSessionResponse);
```
jui8wang, line 83: how would only "ssl handshake complete" can call "whole the session creation complete", shouldn't u wait for the server greeting? u have 2 factors to call it done.  Dont forget.

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
```diff
@@ -0,0 +1,319 @@
...
+            }
+            break;
+
+        case GET_STARTTLS_RESP:
+            // receive startTls response, remove this handler and add SslHandler
+            pipeline.remove(this);
+
+            if (serverResponse.getCode().value() == SmtpResponse.Code.GREETING) { // successful response
+                // add the command response handler
+                final SmtpAsyncSessionImpl session = new SmtpAsyncSessionImpl(ctx.channel(), this.logger, this.logOpt, this.sessionId, pipeline,
```
jui8wang, line 231: remove this.  U should not create session now.  U should create it only when server greeting is obtained.

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
```diff
@@ -0,0 +1,319 @@
...
+                        this.sessionCtx);
+                final SmtpAsyncCreateSessionResponse createSessionResponse = new SmtpAsyncCreateSessionResponse(session, serverResponse);
```
jui8wang, line 233: same.  please remove

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
```diff
@@ -104,75 +117,133 @@ public SmtpAsyncClient(final int numThreads) throws SSLException {
...
+                } else if (isStarttls) {
+                    sessionMode = SessionMode.PLAIN_STARTTLS;
+                } else {
+                    sessionMode = SessionMode.NON_SSL;
+                }
+
+                // add session specific handlers
+                switch (sessionMode) {
+                case SSL_WITH_STARTTLS:
+                    pipeline.addFirst(SslDetectHandler.HANDLER_NAME, new SslDetectHandler(sessionCount.get(), sessionData, config,
```
jui8wang, line 209: dont addFirst twice.  when use addFirst, that one should be absolute first.   U can use addAfter() to insert that after SSLHandler

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
cifer828: sure

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
cifer828: this listener will only be added when we have already received server greeting, in line 229.  the only response from server after we add SslHandler is the ServerHello which will be digested by SslHandler. so i think as long as the sslConnectionFuture is success, whole the session creation completes.

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
cifer828: sure.

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
cifer828: sure

core/src/main/java/com/yahoo/smtpnio/async/netty/SslHandlerBuilder.java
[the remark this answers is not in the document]
cifer828: as discussed, a package static method createSSLHandler() will be used to create sslHandlers

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
cifer828: moved SmtpAsyncClient and SslDetectHandler to the same package and make createStarttlsSession() to package level

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
cifer828: as discussed, all logger.error() are combined into one in handleSessionFailed()

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
```diff
@@ -247,16 +259,54 @@ public long applyAsLong(final long counter) { // increment by 1 unless it overfl
...
-     * @param channel the channel
+     * @param sessionId client's session id
+     * @param sessionData connection data for the session
+     * @param ex exception that causes the session failure
+     * @param sessionCreatedFuture SMTP session future
+     * @param channel session channel to close
+     * @param isSsl whether session use ssl or not
      */
-    private void closeChannel(@Nullable final Channel channel) {
+    private void handleSessionFailed(@Nonnull final Object sessionId, @Nonnull final SmtpAsyncSessionData sessionData,
```
cifer828, line 271: this method is used to combine code

core/src/main/java/com/yahoo/smtpnio/async/client/StarttlsHandler.java
```diff
@@ -205,19 +203,14 @@ public void decode(@Nonnull final ChannelHandlerContext ctx, @Nonnull final Smtp
             break;
 
         case GET_HELO_RESP:
-            // check HELO responses
```
cifer828: remove checking STARTTLS capability in HELO response. according to https://tools.ietf.org/html/rfc5321#section-3.2, HELO response should be different from EHLO and shouldn't contain extensions.

core/src/main/java/com/yahoo/smtpnio/async/client/StarttlsHandler.java
```diff
@@ -0,0 +1,304 @@
...
+            break;
+
+        default:
+            // shouldn't reach here
+            ctx.fireExceptionCaught(new SmtpAsyncClientException(FailureType.STARTTLS_FALIED, "Invalid state"));
+        }
+
+        // close channel and clean up if startTls failed
+        if (isSessionFutureFailed) {
+            handleSessionFailed(ctx,
```
jui8wang, line 245: add a comment that pipeline handlers will all be removed thru handleSessionFailed's close.

core/src/main/java/com/yahoo/smtpnio/async/client/StarttlsHandler.java
```diff
@@ -0,0 +1,304 @@
...
+        handleSessionFailed(ctx, new SmtpAsyncClientException(FailureType.CONNECTION_FAILED_EXCEPTION, cause, this.sessionId, this.sessionCtx));
+    }
+
+    @Override
+    public void userEventTriggered(@Nonnull final ChannelHandlerContext ctx, @Nonnull final Object msg) {
+        if (msg instanceof IdleStateEvent) { // Handle the IdleState if needed
+            final IdleStateEvent event = (IdleStateEvent) msg;
+            if (event.state() == IdleState.READER_IDLE) {
+                handleSessionFailed(ctx,
+                        new SmtpAsyncClientException(FailureType.CONNECTION_FAILED_EXCEED_IDLE_MAX, this.sessionId, this.sessionCtx));
```
jui8wang, line 261: remember not to use CONNECTION related failure type, since u already connect.

core/src/main/java/com/yahoo/smtpnio/async/client/StarttlsHandler.java
```diff
@@ -0,0 +1,304 @@
...
+                    isSessionFutureFailed = true;
+                }
+            }
+            break;
+
+        case GET_STARTTLS_RESP:
+            // receive startTls response, remove this handler and add SslHandler
+            pipeline.remove(this);
+
+            if (serverResponse.getCode().value() == SmtpResponse.Code.GREETING) { // successful response
```
jui8wang, line 222: check if it is last line

jui8wang, line 222: as discussed, lets not reuse greeting's success code, lets define a Code.STARTTLS_SUCCESS code there.

core/src/main/java/com/yahoo/smtpnio/async/client/StarttlsHandler.java
```diff
@@ -0,0 +1,304 @@
...
+        this.sessionData = sessionData;
+        this.startTlsState = StartTlsState.GET_SERVER_GREETING;
+        this.receivedStarttlsCapability = false;
+    }
+
+    @Override
+    public void decode(@Nonnull final ChannelHandlerContext ctx, @Nonnull final SmtpResponse serverResponse, @Nonnull final List<Object> out) {
+        final Channel channel = ctx.channel();
+        final ChannelPipeline pipeline = ctx.pipeline();
+        boolean isSessionFutureFailed = false; // the session future should be failed in this call
```
jui8wang, line 169: change it to Exception failedCause = null;

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
```diff
@@ -104,88 +119,194 @@ public SmtpAsyncClient(final int numThreads) throws SSLException {
...
+                    sessionMode = SessionMode.NON_SSL;
+                }
+
+                final ChannelPipeline pipeline = ch.pipeline();
+
+                // add session specific handlers
+                switch (sessionMode) {
+                case SSL_WITH_STARTTLS:
+                    pipeline.addFirst(SSL_HANDLER, sslHandler);
+                    pipeline.addAfter(SSL_HANDLER, SslDetectHandler.HANDLER_NAME, new SslDetectHandler(sessionCount.get(), sessionData, config,
```
jui8wang, line 224: why do u need to pass logger from caller?  why not initialize in its own constructor

core/src/main/java/com/yahoo/smtpnio/async/client/StarttlsHandler.java
cifer828: changed to CHANNEL_TIMEOUT

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
```diff
@@ -104,88 +119,194 @@ public SmtpAsyncClient(final int numThreads) throws SSLException {
...
+                            LoggerFactory.getLogger(SslDetectHandler.class), debugOption, smtpAsyncClient, sessionCreatedFuture));
+                    pipeline.addLast(SmtpClientConnectHandler.HANDLER_NAME, new SmtpClientConnectHandler(sessionCreatedFuture,
+                            LoggerFactory.getLogger(SmtpClientConnectHandler.class), debugOption, sessionId, sessionCtx));
```
jui8wang, line 227: same comment regarding the logger

core/src/main/java/com/yahoo/smtpnio/async/client/StarttlsHandler.java
cifer828: sure

cifer828: sure

core/src/main/java/com/yahoo/smtpnio/async/client/StarttlsHandler.java
cifer828: changed to SmtpAsyncClientException failedCause = null

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
jui8wang: it is ok not to cover

core/src/main/java/com/yahoo/smtpnio/async/client/StarttlsHandler.java
```diff
@@ -0,0 +1,304 @@
...
+            // check EHLO responses
+            if (serverResponse.getMessage().equalsIgnoreCase(STARTTLS)) {
+                this.receivedStarttlsCapability = true;
+            }
+
+            // decide next state when we receive last response
+            if (serverResponse.isLastLineResponse()) {
+                if (serverResponse.getCode().value() != Code.EHLO_SUCCESS) {
+                    // EHLO fails, try HELO as fallback
+                    channel.writeAndFlush(new HelloCommand(EHLO_CLIENT_NAME).getCommandLineBytes());
```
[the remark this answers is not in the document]
cifer828: remove helo state

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
```diff
@@ -104,88 +119,194 @@ public SmtpAsyncClient(final int numThreads) throws SSLException {
...
+                    break;
+
+                default:
+                    // shouldn't reach here
+                    throw new SmtpAsyncClientException(FailureType.INVALID_INPUT, "Invalid state");
+                }
+
                 if (logger.isTraceEnabled() || debugOption == SmtpAsyncSession.DebugMode.DEBUG_ON) {
-                    logger.debug(CONNECT_RESULT_REC, sessionId, sessionCtx, "success",  host, port, enableSsl, sniNames);
+                    logger.debug(CONNECT_RESULT_REC, sessionId, sessionCtx, "success", host, port, isSsl, sniNames);
```
jui8wang, line 255: could u add the sessionMode in the log

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
```diff
@@ -46,14 +47,31 @@
...
+    public static final String CONNECT_RESULT_REC = "[{},{}] connect operation complete. result={}, host={}, port={}, sslEnabled={}, sniNames={}";
+
+    /** Debug record string template for reconnection. */
+    public static final String RE_CONNECTION_REC = "[{},{}] try starttls. Re-connecting with non-ssl. "
+            + "host={}, port={}, sslEnabled={}, startTlsEnabled={}, sniNames={}";
 
-    /** The SSL context. */
-    private final SslContext sslContext;
+    /** Handler name for the ssl handler. */
+    static final String SSL_HANDLER = "sslHandler";
```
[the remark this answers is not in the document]
cifer828: StarttlsHandler also uses this constant, i made it private and add the same constant in StarttlsHandler

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
cifer828: sure

core/src/main/java/com/yahoo/smtpnio/async/client/StarttlsHandler.java
```diff
@@ -0,0 +1,304 @@
...
+                channel.writeAndFlush(new ExtendedHelloCommand(EHLO_CLIENT_NAME).getCommandLineBytes());
+                this.startTlsState = StartTlsState.GET_EHLO_RESP;
+            } else {
+                isSessionFutureFailed = true;
+            }
+            break;
+
+        case GET_EHLO_RESP:
+            // check EHLO responses
+            if (serverResponse.getMessage().equalsIgnoreCase(STARTTLS)) {
```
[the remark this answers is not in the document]
cifer828: sure

core/src/main/java/com/yahoo/smtpnio/async/client/StarttlsHandler.java
```diff
@@ -0,0 +1,304 @@
...
+            }
+        }
+    }
+
+    @Override
+    public void channelInactive(@Nonnull final ChannelHandlerContext ctx) {
+        if (this.sessionCreatedFuture == null) {
+            return; // cleanup() has been called, leave
+        }
+        handleSessionFailed(ctx, new SmtpAsyncClientException(FailureType.CONNECTION_INACTIVE));
```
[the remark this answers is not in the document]
cifer828: create new failure type CHANNEL_INACTIVE. also create other new types for different failures.

core/src/main/java/com/yahoo/smtpnio/async/client/StarttlsHandler.java
```diff
@@ -0,0 +1,310 @@
...
+            break;
+
+        default:
+            // shouldn't reach here
+            ctx.fireExceptionCaught(new SmtpAsyncClientException(FailureType.STARTTLS_FALIED, "Invalid state"));
+        }
+
+        // close channel and clean up if startTls failed
+        if (isSessionFutureFailed) {
+            this.logger.error("[{},{}] startTls failed, server response: {}", this.sessionId, this.sessionCtx, serverResponse.toString());
```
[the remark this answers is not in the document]
cifer828: moved

core/src/main/java/com/yahoo/smtpnio/async/client/SslDetectHandler.java
```diff
@@ -0,0 +1,141 @@
...
+        final Collection<String> sniNames = this.sessionData.getSniNames();
+        final Object sessionCtx = this.sessionData.getSessionContext();
+
+        // ssl failed, re-connect with plain connection
+        if (cause.getCause() instanceof NotSslRecordException) {
+            close(ctx);
+            // if startTls is enabled, try to create a new connection without ssl
+            this.smtpAsyncClient.createStarttlsSession(this.sessionData, this.sessionConfig, this.logOpt, this.sessionCreatedFuture);
+            if (this.logger.isTraceEnabled() || this.logOpt == SmtpAsyncSession.DebugMode.DEBUG_ON) {
+                this.logger.debug(SSL_DETECT_REC, this.sessionId, sessionCtx, "Not available", host, port, true, sniNames);
```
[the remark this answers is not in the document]
cifer828: it uses twice. write literal for both of them.

core/src/main/java/com/yahoo/smtpnio/async/client/StarttlsHandler.java
```diff
@@ -0,0 +1,304 @@
...
+                this.startTlsState = StartTlsState.GET_EHLO_RESP;
+            } else {
+                isSessionFutureFailed = true;
+            }
+            break;
+
+        case GET_EHLO_RESP:
+            // check EHLO responses
+            if (serverResponse.getMessage().equalsIgnoreCase(STARTTLS)) {
+                this.receivedStarttlsCapability = true;
```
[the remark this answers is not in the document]
cifer828: don't think so. decode() will be called several times with this state because EHLO responses will be split into multiple lines and each line will be digested as a SmtpResponse. if receivedStarttlsCapability is local, when we receive STARTTLS in the middle, the startTlsState will switch. then next time when decode() is called, it's in the wrong state (expect GET_EHLO_RESP, but has switched to GET_STARTTLS_RESP)

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
cifer828: initialize logger in a package constructor for the sake of logger verification in UT

core/src/main/java/com/yahoo/smtpnio/async/client/StarttlsHandler.java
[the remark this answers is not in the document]
cifer828: sure

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
```diff
@@ -104,88 +119,194 @@ public SmtpAsyncClient(final int numThreads) throws SSLException {
...
-                        final SSLEngine engine = sslContext.newEngine(ch.alloc(), host, port);
-                        engine.setSSLParameters(params);
-                        pipeline.addFirst(SSL_HANDLER, new SslHandler(engine)); // in / outbound
-                    } else {
-                        pipeline.addFirst(SSL_HANDLER, sslContext.newHandler(ch.alloc(), host, port));
-                    }
-                }
 
+                // avoid negative ids
                 final long sessionId = sessionCount.getAndUpdate(new LongUnaryOperator() { // atomic update
```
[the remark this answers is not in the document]
cifer828: this lambda expression can't pass style check, not sure why. i refer to imap-nio's [ImapAsyncClient](https://github.com/yahoo/smtp-client-nio/blob/master/core/src/main/java/com/yahoo/smtpnio/async/exception/SmtpAsyncClientException.java), and use incrementAndGet() with compareAndSet(Long.MAX_VALUE - 1, 1) instead.

core/src/main/java/com/yahoo/smtpnio/async/client/StarttlsHandler.java
```diff
@@ -0,0 +1,304 @@
...
+                    isSessionFutureFailed = true;
+                }
+            } else {
+                isSessionFutureFailed = true;
+            }
+            break;
+
+        default:
+            // shouldn't reach here
+            ctx.fireExceptionCaught(new SmtpAsyncClientException(FailureType.STARTTLS_FALIED, "Invalid state"));
```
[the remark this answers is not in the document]
cifer828: replace the flag by failedCause exception

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncSessionConfig.java
```diff
@@ -21,6 +21,9 @@
     /** Maximum time in milliseconds for read timeout. The maximum time allowing no responses from server since client command sent. */
     private int readTimeoutMillis = DEFAULT_READ_TIMEOUT_MILLIS;
 
+    /** Whether to enable Starttls for this client. */
+    private boolean enableStarttls = true;
```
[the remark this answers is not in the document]
cifer828: changed to false

core/src/main/java/com/yahoo/smtpnio/async/client/SslDetectHandler.java
```diff
@@ -57,14 +53,31 @@
...
+     *
+     * @param sessionId client's session id
+     * @param sessionData connection data for this session
+     * @param sessionConfig configurations for this session
+     * @param logger logger for debugging messages
+     * @param logOpt debugging options
+     * @param smtpAsyncClient Client that holds this connection
+     * @param sessionCreatedFuture SMTP session future
+     */
+    SslDetectHandler(final long sessionId, @Nonnull final SmtpAsyncSessionData sessionData, @Nonnull final SmtpAsyncSessionConfig sessionConfig,
```
cifer828, line 78: create this package constructor for the sake of verifying logs in UTs.

core/src/test/java/com/yahoo/smtpnio/async/client/SslDetectHandlerTest.java
```diff
@@ -0,0 +1,424 @@
...
+    /**
+     * Tests {@code exceptionCaught} method when other exception is caught.
+     *
+     * @throws IllegalArgumentException will not throw in this test
+     * @throws InterruptedException will not throw in this test
+     * @throws TimeoutException will not throw in this test
+     * @throws IllegalAccessException will not throw in this test
+     */
+    @Test
+    public void testtOtherExceptionCaught() throws InterruptedException, TimeoutException, IllegalArgumentException, IllegalAccessException {
```
fansu, line 326: testt

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
```diff
@@ -104,88 +116,196 @@ public SmtpAsyncClient(final int numThreads) throws SSLException {
...
-                pipeline.addLast(SmtpClientConnectHandler.HANDLER_NAME,
-                        new SmtpClientConnectHandler(sessionFuture, LoggerFactory.getLogger(SmtpAsyncSessionImpl.class),
-                                debugOption, sessionId, sessionCtx));
+
+                // create SslHandler for secure connection
+                SslHandler sslHandler = null;
+                if (isSsl) {
+                    try {
+                        sslHandler = createSSLHandler(ch.alloc(), host, port, sniNames);
+                    } catch (SSLException e) {
```
fansu, line 197: final

core/src/main/java/com/yahoo/smtpnio/async/client/StarttlsHandler.java
```diff
@@ -0,0 +1,302 @@
...
+import io.netty.util.concurrent.Future;
+import io.netty.util.concurrent.GenericFutureListener;
+
+/**
+ * This handler is used to implement startTls flow.
+ *
+ * <p>
+ * Starttls flow consists of four steps:
+ *
+ * <li>1. process server greeting and send EHLO.
```
fansu, line 40: Your UT is failing, can you fix this part? 
https://travis-ci.com/github/yahoo/smtp-client-nio/builds/178681818

fansu:
Can you fix the UTs that showing in job?
https://travis-ci.com/github/yahoo/smtp-client-nio/builds/178681818

core/src/main/java/com/yahoo/smtpnio/async/client/StarttlsHandler.java
cifer828: sure

core/src/main/java/com/yahoo/smtpnio/async/netty/StarttlsHandler.java
cifer828: sure

core/src/main/java/com/yahoo/smtpnio/async/client/SslDetectHandler.java
```diff
@@ -103,6 +109,7 @@ protected void decode(@Nonnull final ChannelHandlerContext ctx, @Nonnull final B
     public void exceptionCaught(@Nonnull final ChannelHandlerContext ctx, @Nonnull final Throwable cause) {
         // ssl failed, re-connect with plain connection
         if (cause.getCause() instanceof NotSslRecordException) {
+            this.isReconnecting = true;
```
nituyahoo, line 112: can we move close and isReconnecting to after the createStartTlsSession

core/src/main/java/com/yahoo/smtpnio/async/client/SslDetectHandler.java
```diff
@@ -131,8 +138,14 @@ private void close(final ChannelHandlerContext ctx) {
 
     @Override
     public void channelInactive(@Nonnull final ChannelHandlerContext ctx) {
-        // after close() has been called on SslHandler failure, channelInactive() will also be called
-        // this method is used to stop channelInactive chain reaching SmtpClientConnectionHandler
+        // only finish the future with exception when it's not re-connecting
+        if (!this.isReconnecting) {
+            if (sessionCreatedFuture == null) {
```
nituyahoo, line 143: do you have unit test for this usecase?

cifer828: yes, line 442 in SslDetectHandlerTest cover this case. i refer to line 338 in SmtpClientConnectionHandlerTest.

core/src/main/java/com/yahoo/smtpnio/async/client/SslDetectHandler.java
cifer828: sure

PR #40, opened 2023-03-27
ISSUE-32 Allow callbacks in SmtpFuture

chibenwa:
This allow avoiding blocking in calling code to unwrap the future
and access the inner value. This empowers reactive/asynchronous/actor
systems integrations.

I chose the callback approach instead of using CompletableFuture in
order to minimize code changes.

## Description

Adds callbacks to SmtpFuture

Similar to https://github.com/yahoo/imapnio/pull/118

## Motivation and Context

Needs to call SmtpFuture.get to do anything usefull with this library however it blocks and defeat any reactive/asynchronous/actor system use of this library.

## How Has This Been Tested?

Unit test. I will also ensure timely that this solves my original issue.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [x] My change requires a change to the documentation.
- [x] I have added tests to cover my changes.

### DIFF
--- a/core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
+++ b/core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
@@ -118,7 +118,7 @@ public class SmtpAsyncClient {
      * @param debugOption the debugging option used
      * @return the future containing the result of the request
      */
-    public SmtpFuture<SmtpAsyncCreateSessionResponse> createSession(@Nonnull final SmtpAsyncSessionData sessionData,
+    public Future<SmtpAsyncCreateSessionResponse> createSession(@Nonnull final SmtpAsyncSessionData sessionData,
             @Nonnull final SmtpAsyncSessionConfig config, @Nonnull final SmtpAsyncSession.DebugMode debugOption) {
         bootstrap.option(ChannelOption.CONNECT_TIMEOUT_MILLIS, config.getConnectionTimeout());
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
```diff
@@ -118,7 +118,7 @@ public SmtpAsyncClient(final int numThreads) {
      * @param debugOption the debugging option used
      * @return the future containing the result of the request
      */
-    public Future<SmtpAsyncCreateSessionResponse> createSession(@Nonnull final SmtpAsyncSessionData sessionData,
+    public SmtpFuture<SmtpAsyncCreateSessionResponse> createSession(@Nonnull final SmtpAsyncSessionData sessionData,
```
fansu, line 121: Revert this line and keep same Future signature for now. So this Pull Request will only has new callbacks features added with 100% backward compatibility.

fansu (changes requested):
Please address one comment, otherwise looks good.

core/src/main/java/com/yahoo/smtpnio/async/client/SmtpAsyncClient.java
chibenwa: This would force async user to cast in their code

SmtpFuture is a subclass of future hence compatibility is not an issue.

Btw similar changes were accepted in imapnio...

fansu (approved):

PR #60, opened 2026-06-24
Replace nexus-staging-maven-plugin with central-publishing-maven-plugin

sundarms:
Replace nexus-staging-maven-plugin with central-publishing-maven-plugin

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -582,15 +582,13 @@
                         </executions>
                     </plugin>
                     <plugin>
-                        <groupId>org.sonatype.plugins</groupId>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -5,7 +5,7 @@
     <groupId>com.yahoo.smtpnio</groupId>
     <artifactId>smtpnio</artifactId>
     <packaging>pom</packaging>
-    <version>1.1.5</version>
+    <version>1.1.6</version>
```
fansu, line 8: version upgrade for core/pom.xml too

fansu (approved):

PR #62, opened 2026-09-18
Validate command arguments against RFC 5321 section 4.1.2

fansu:
Add SmtpArgumentFormatter and write every argument through it, into a
ByteBuf for the command line or a StringBuilder for the SASL payloads the
AUTH commands assemble before base64. Arguments are confined to printable
US-ASCII, %d32-126: section 4.1.2 says non-ASCII and ASCII control
characters "MUST NOT be used in MAIL or RCPT commands", and neither
qtextSMTP nor quoted-pairSMTP reaches past %d126, so there is nothing to
encode and refusing to build the command is the only option. Refusing the
control half is what stops CR and LF terminating the line, and NUL and
SOH injecting a field into the AUTH PLAIN and XOAUTH2 payloads. Refusing
the non-ASCII half ends a silent corruption: those characters used to
reach getBytes(US_ASCII) and flatten to '?', sending mail to an address
the caller never asked for.

Mail parameters get their own productions, esmtp-keyword and esmtp-value.
A space in a value would otherwise begin a further esmtp-param and an
equals sign would read as the keyword separator.

Besides the four classes named in the report this also covers
MailCommand.MailParameter and the credentials of all three AUTH
mechanisms, which had the same gap.

Arguments are checked while the command line is built rather than in the
constructors. SmtpAsyncClientException is checked and SmtpRequest already
declares it on getCommandLineBytes, so throwing from the constructors
would have forced a throws clause onto every one of them and broken their
call sites.

Behaviour changes, though binary and source compatible: CRLF that
previously passed through and silently sent two commands now raises
SmtpAsyncClientException with FailureType.INVALID_INPUT, as does a
non-ASCII address or a parameter keyword outside the production.
MailCommandTest asserted the latter, an underscore in "Just_key", while
claiming to follow the RFC; it now uses a hyphen.

Bump the version to 1.1.8 and correct the README install snippet, which
still advertised 1.1.5.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

## Checklist:

- [x] My change requires a change to the documentation.
- [x] I have added tests to cover my changes.


core/src/main/java/com/yahoo/smtpnio/async/request/AbstractSmtpCommand.java
```diff
@@ -34,6 +40,36 @@ protected AbstractSmtpCommand(@Nonnull final String command) {
...
+     * rather than silently forwarded.
+     * </p>
+     *
+     * @param value the argument to validate
+     * @param name the name of the argument, used in the error message
+     * @return the argument, unchanged, when it is safe to write to the wire
+     * @throws IllegalArgumentException when the argument contains a control character
+     */
+    @Nonnull
+    protected static String validateArgument(@Nonnull final String value, @Nonnull final String name) {
```
sundarms: Can we add UT for this?

fansu: Done, added.

core/src/test/java/com/yahoo/smtpnio/async/request/CommandArgumentValidationTest.java
```diff
@@ -0,0 +1,164 @@
+/*
+ * Copyright Verizon Media
```
sundarms: Not a blocker but we should revisit fix this later across all files.

sundarms (approved):

core/src/test/java/com/yahoo/smtpnio/async/request/CommandArgumentValidationTest.java
fansu: ok

core/src/main/java/com/yahoo/smtpnio/async/request/SmtpArgumentFormatter.java
```diff
@@ -0,0 +1,89 @@
...
+     * @param src the argument to inspect
+     * @param name the name of the argument, used in the error message
+     * @throws SmtpAsyncClientException when the argument contains a control character
+     */
+    private void ensureNoControlChar(@Nonnull final String src, @Nonnull final String name) throws SmtpAsyncClientException {
+        for (int i = 0; i < src.length(); i++) {
+            final char c = src.charAt(i);
+            if (c <= LAST_C0_CONTROL_CHAR || c == DEL_CHAR) {
+                // the offending value is deliberately excluded from the message, it may hold a credential
+                throw new SmtpAsyncClientException(SmtpAsyncClientException.FailureType.INVALID_INPUT,
```
nvsnvikram: Please use smtp rfc to validate: https://www.rfc-editor.org/rfc/rfc5321.html#section-4.1.2.

fansu: Fix the code according to rfc5321.

nvsnvikram (approved):

